PR #58, opened 2020-11-12
Propose Authentication allowed on DID resolution

jose-gataca:


TomCJones:
This seems to me like a variant of getting private information out of the did doc. This problem was well articulated by Elizabeth Reneris quite awhile ago. This pull request seems to solve that, but basically puts an end to the idea of public did docs.  I think that this is a large issue with resolution that is no where near a solution.

jose-gataca:
I don't think it puts and end to the idea of public DID docs, but opens the door to having public and private DID docs. If DIDs are considered as resources, those can be both public or private on the internet.

On the other hand, I don't find the capability of requiring Authentication authentication similar to getting private information of the DID doc. The information of the DID document should be always the same.

The main goal behind the Authentication for resolution is however not enabling private DIDs, but public-permissionned resolution. It could be possible that Authentication is not the best solution with that goal, however, it seems the discussion is there.

PR #63, opened 2021-04-01
Update DID Resolution contexts to include metadata properties.

peacekeeper:
This adds known DID resolution and DID document metadata properties to the JSON-LD context.

### DIFF
--- a/contexts/did-resolution-v1.json
+++ b/contexts/did-resolution-v1.json
@@ -5,40 +5,39 @@
     "id": "@id",
     "type": "@type",
     "xsd": "http://www.w3.org/2001/XMLSchema#",
-    "didres": "https://w3id.org/did-resolution#",
 
-    "didDocument": "didres:didDocument",
+    "didDocument": "https://w3id.org/did-resolution#didDocument",
 
-    "didResolutionMetadata": "didres:didResolutionMetadata",
-    "contentType": "didres:contentType",
-    "error": "didres:error",
+    "didResolutionMetadata": "https://w3id.org/did-resolution#didResolutionMetadata",
+    "contentType": "https://w3id.org/did-resolution#contentType",
+    "error": "https://w3id.org/did-resolution#error",
 
-    "didDocumentMetadata": "didres:didDocumentMetadata",
+    "didDocumentMetadata": "https://w3id.org/did-resolution#didDocumentMetadata",
     "created": {
-      "@id": "didres:created",
+      "@id": "https://w3id.org/did-resolution#created",
       "@type": "xsd:dateTime"
     },
     "updated": {
-      "@id": "didres:updated",
+      "@id": "https://w3id.org/did-resolution#updated",
       "@type": "xsd:dateTime"
     },
     "deactivated": {
-      "@id": "didres:deactivated",
+      "@id": "https://w3id.org/did-resolution#deactivated",
       "@type": "xsd:boolean"
     },
     "nextUpdate": {
-      "@id": "didres:nextUpdate",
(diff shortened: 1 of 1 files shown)


contexts/did-resolution-v1.json
```diff
@@ -1,22 +1,44 @@
...
-    {
-      "@version": 1.1
+  "@context": {
+    "@version": 1.1,
+    "id": "@id",
+    "type": "@type",
+    "xsd": "http://www.w3.org/2001/XMLSchema#",
+    "didres": "https://w3id.org/did-resolution#",
+
+    "didDocument": "didres:didDocument",
```
kdenhartog, line 9: can we fully qualify rather than use the short-form? That's becoming a standard we've started using elsewhere that's worked pretty well.

contexts/did-resolution-v1.json
```diff
@@ -1,22 +1,44 @@
 {
-  "@context": [
-    {
-      "@version": 1.1
+  "@context": {
+    "@version": 1.1,
```
kdenhartog, line 3: Can we add `@protected` here?

contexts/did-resolution-v1.json
kdenhartog, line 9: Would it make sense to make the IRI defined here set to the URL to the DID Document context?

peacekeeper: Yes sounds good.

contexts/did-resolution-v1.json
peacekeeper: Yes let's do that.

contexts/did-resolution-v1.json
peacekeeper: Hmm interesting..! I guess this would mean that the `didDocument` term here in the resolution result would then always be linked to a certain version of the DID document context, but I think that might be a good thing.

kdenhartog: Ohh good point about linking to specific versions. In some cases I can see it as a good thing and in others I could see it as a pain. It really depends on how we version. E.g. a minor version update should keep the same IRI, but for a v2 do we add an additional property with a different IRI (e.g. "didDocumentV2": "https://w3.org/ns/did/v2")

peacekeeper:
@kdenhartog I applied all your suggestions.

contexts/did-resolution-v1.json
```diff
@@ -1,22 +1,44 @@
...
+    "@protected": true,
+    "id": "@id",
+    "type": "@type",
+    "xsd": "http://www.w3.org/2001/XMLSchema#",
```
kdenhartog, line 7: nitpick: we could drop this alias here.

kdenhartog (approved):
I'm good to approve with our without shortner. Looks good to me.

peacekeeper:
> I'm good to approve with our without shortner. Looks good to me.

Ah I misunderstood, I thought you only meant removing the `didres:` prefix. Hmm in the case of `xsd:` I find this more readable... So I think I'll merge as-is, unless there's some good reason for not using prefixes at all.

kdenhartog:
> 
> 
> > I'm good to approve with our without shortner. Looks good to me.
> 
> Ah I misunderstood, I thought you only meant removing the `didres:` prefix. Hmm in the case of `xsd:` I find this more readable... So I think I'll merge as-is, unless there's some good reason for not using prefixes at all.

Works for me. The main reason I like removing them for IRIs is to make it clear what the exact IRI is. For the type it's not as big of a deal in my opinion. Just mentioned for consistency sake with other contexts.

Issue #65, opened 2021-07-12
What's the expected behavior for a DID Document that contains the same id property for publicKeys?


kdenhartog:
Given the following DID Document, what's the expected result from the dereference function?

```json
  {
    "@context": [
      "https://www.w3.org/ns/did/v1",
      "https://w3id.org/security/suites/ed25519-2020/v1"
    ],
    "id": "did:example:123",
    "verificationMethod": [
      {
        "id": "did:example:123#publicKey",
        "type": "Ed25519VerificationKey2020",
        "controller": "did:example:123",
        "publicKeyMultibase": "zAKJP3f7BD6W4iWEQ9jwndVTCBq8ua2Utt8EEjJ6Vxsf"
      },
      {
        "id": "#publicKey",
        "type": "Ed25519VerificationKey2020",
        "controller": "did:example:123",
        "publicKeyMultibase": "differentPublicKeyExample....."
      }
    ],
}
```

`dereference("did:example:123#publicKey", {"accept": "application/did+ld+json"})` => ????

peacekeeper:
Good question, I think that's basically a question for JSON-LD itself and is not really specific to DIDs or DID documents or DID URL dereferencing.

The JSON-LD spec at https://w3c.github.io/json-ld-syntax/#application-ld-json says this:

> Fragment identifiers used with application/ld+json are treated as in RDF syntaxes, as per RDF 1.1 Concepts and Abstract Syntax [RDF11-CONCEPTS].

But this isn't very helpful, is it?

My assumption has been that the dereferencer picks the JSON object with the matching `id`, and also adds all applicable contexts, i.e. the result in your specific example would be:

```
  {
  	"@context": [
  		"https://www.w3.org/ns/did/v1",
  		"https://w3id.org/security/suites/ed25519-2020/v1"
  	],
  	"id": "did:example:123#publicKey",
  	"type": "Ed25519VerificationKey2020",
  	"controller": "did:example:123",
  	"publicKeyMultibase": "zAKJP3f7BD6W4iWEQ9jwndVTCBq8ua2Utt8EEjJ6Vxsf"
  }
```

But this isn't really documented anywhere; and for `application/did+json` the use of the fragment is completely undefined (probably assumed to be the same as for `application/did+ld+json`).

kdenhartog:
[...]

PR #67, opened 2021-08-14
update clarification around handling of conflicting id's in json-ld

kdenhartog:
This only addresses issue #65 for JSON-LD. I didn't spot an easy way to slot it in for other formats - nor was I sure if it made sense to process these differently. I figured it goes in the right direction though and was worth adding something in at least.

### DIFF
--- a/index.html
+++ b/index.html
@@ -643,7 +643,9 @@ dereference ( did-url, did-url-dereferencing-input-metadata ) <br>
 				<a>DID document</a>. This is called the <var>output resource</var>.
 				When selecting the JSON-LD object from the <a>DID document</a>, the absolute
 				<a>DID URL</a> used to identify a graph node MUST be unique and present only once in the
-				<a>DID document</a>. If the identifer of the graph node is not unique, an error MUST be thrown.
+				<a>DID document</a>. "If the identifier of the graph node is not unique,
+				including if a relative or base IRI mapped to an absolute IRI collides with
+				a different graph node's absolute IRI, then an error MUST be thrown."
 				<li>Return the <var>output resource</var>.
 				<div class="issue" data-number="9">
 				<p>Mention relative IRIs and that the DID itself is considered the base IRI for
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -641,8 +641,9 @@ <h2>Dereferencing the Secondary Resource</h2>
 				<li>From the <var>resolved <a>DID document</a></var>, select the JSON-LD object whose <code>id</code>
 				property matches the <var>input <a>DID URL</a></var>, e.g. a public key or service endpoint in the
 				<a>DID document</a>. This is called the <var>output resource</var>.
-				Selecting the JSON-LD object from the <a>DID document</a> MUST take into account relative IRIs and
-				the base IRI of the <var>resolved <a>DID document</a></var>.</li>
+				When selecting the JSON-LD object from the <a>DID document</a>, the absolute
+				<a>DID URL</a> used to identify a graph node MUST be unique and present only once in the
+				<a>DID document</a>. If the indentifer of the graph node is not unique, an error MUST be thrown.
```
TallTed, line 646:
```suggestion
				property matches the <var>input <a>DID URL</a></var>, e.g., a public key or service endpoint in the
				<a>DID document</a>. This is called the <var>output resource</var>.

				When selecting the JSON-LD object from the <a>DID document</a>, the absolute
				<a>DID URL</a> used to identify a graph node MUST be unique and present only once in the
				<a>DID document</a>. If the identifier of the graph node is not unique, an error MUST be thrown.
```

kdenhartog: I'm not able to figure out what this suggestion is doing. The only difference I see here is that line 644 now has a space between the two. Given these are in a `<li>` HTML format I'm not sure what this adds. Can you clarify the suggestion @TallTed ?

kdenhartog:
> > I think this would make it more clear about the specifics of what needs to be checked for and how it should be handled.
> 
> Yes I think that sounds better.. If you update the PR we can merge that..

@peacekeeper Updated this now with the text we agreed on above. Thanks for the review

peacekeeper (approved):

PR #68, opened 2021-08-23
Prevent double fragment in service endpoint URL.

clehner:
This adds a check to ensure that an *output service endpoint URL* does not have two fragment components. This is equivalent to the requirement in [Service Endpoint Construction](https://w3c-ccg.github.io/did-resolution/#input) that "the input DID URL and input service endpoint URL MUST NOT both have a fragment component". The reason to add it here is because when the Service Endpoint Construction algorithm is called (in Step 1.2 of [Dereferencing the Primary Resource](https://w3c-ccg.github.io/did-resolution/#dereferencing-algorithm-primary)), the *input DID URL* has already had its fragment removed (in Step 2 of the [DID URL dereferencing algorithm](https://w3c-ccg.github.io/did-resolution/#dereferencing), so the existing check is not effectual. The removed fragment is appended to the *output service endpoint URL* in [Dereferencing the Secondary Resource](https://w3c-ccg.github.io/did-resolution/#dereferencing-algorithm-secondary) (Step 2.1), so this PR adds the check there.

### DIFF
--- a/index.html
+++ b/index.html
@@ -480,6 +480,8 @@ dereference ( didUrl, dereferenceOptions ) <br>
 			and the <var>input <a>DID URL</a></var> contains a <a>DID fragment</a>:
 			<pre class="example nohighlight">did:example:1234?service=files&relativeRef=%2Fmyresume%2Fdoc%3Fversion%3Dlatest#intro</pre>
 			<ol class="algorithm">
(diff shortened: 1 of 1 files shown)


kdenhartog (approved):

dmitrizagidulin (approved):
:+1:

PR #70, opened 2022-02-18
Add error code for internalError

aljones15:
This adds the error `internalError` with a suggested status code of 500.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1130,7 +1130,7 @@ https://example.com/messages/8377464/some/path?query#frag
 	<h2>internalError</h2>
     <p>When an unexpected error occurs during did resolution or dereferencing 
     the resolution or dereferencing meta data error should be <i>internalError</i> and status code 500 
-    should be returned.</p>
+    should be returned when an HTTP(s) binding is in use.</p>
 </section>
 
 <section id="security-privacy-considerations">
--- a/index.html
+++ b/index.html
@@ -1128,6 +1128,10 @@ https://example.com/messages/8377464/some/path?query#frag
 <section id="errors">
 	<h1>Errors</h1>
 	<p class="issue">Do we need to define a list of error conditions, codes, etc.</p>
+    <h2>internalError</h2>
+    <p>When an unexpected error occurs during did resolution or dereferencing 
+    the resultion or dereferencing meta data error should be <i>internalError</i> and status code 500 
+    should be returned.</p>
 
 </section>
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1127,8 +1127,10 @@ <h2>Example</h2>
 
 <section id="errors">
 	<h1>Errors</h1>
-	<p class="issue">Do we need to define a list of error conditions, codes, etc.</p>
-
+	<h2>internalError</h2>
+    <p>When an unexpected error occurs during did resolution or dereferencing 
+    the resolution or dereferencing meta data error should be <i>internalError</i> and status code 500 
+    should be returned.</p>
```
kdenhartog, line 1133:
```suggestion
    should be returned when an HTTP(s) binding is in use.</p>
```

kdenhartog (approved):
One minor nit suggested in order to clarify that the suggestion is for that particular binding. It may make sense for bindings to define these aspects rather than the error section, but for now this is good enough in my opinion.

peacekeeper:
I agree with @kdenhartog that since DID Resolution doesn't necessarily use HTTP, HTTP status codes should be defined in the section about the HTTP(S) Binding:
https://w3c-ccg.github.io/did-resolution/#bindings-https

E.g. that section says if the error is "notFound", then return HTTP status code 404.

But I also agree we could merge this now and then improve later.

peacekeeper (approved):

aljones15:
@peacekeeper @kdenhartog I can't merge this PR, but I did take the suggestion. So let me know if this is good to go and thanks for the feedback.

peacekeeper:
Looks good, thanks.

index.html
```diff
@@ -1127,8 +1127,10 @@ <h2>Example</h2>
 
 <section id="errors">
 	<h1>Errors</h1>
-	<p class="issue">Do we need to define a list of error conditions, codes, etc.</p>
-
+	<h2>internalError</h2>
+    <p>When an unexpected error occurs during did resolution or dereferencing 
+    the resolution or dereferencing meta data error should be <i>internalError</i> and status code 500
```
TallTed, line 1132: Got merged too quick...  Needs 2 commas, as shown --
```suggestion
    <p>When an unexpected error occurs during did resolution or dereferencing, 
    the resolution or dereferencing meta data error should be <i>internalError</i>, and status code 500 
```

peacekeeper: Sorry I should have waited longer. I now applied your suggestion in https://github.com/w3c-ccg/did-resolution/commit/746b326570895fc6ce014a659e5d9d55ba9efec0.

PR #71, opened 2022-07-27
Add did key errors to spec.

aljones15:
Adds errors from the [did key spec](https://w3c-ccg.github.io/did-method-key/) to the did resolution.

Errors added:

- invalidPublicKeyLength
- invalidPublicKey
- invalidDidUrl
- unsupportedPublicKeyType
- invalidPublicKeyType

### DIFF
--- a/index.html
+++ b/index.html
@@ -944,6 +944,22 @@ https://example.com/messages/8377464/some/path?query#frag
 	<h2>internalError</h2>
 	<p>When an unexpected error occurs during <a>DID Resolution</a> or <a>DID URL dereferencing</a>,
 		the value of the DID Resolution or DID URL Dereferencing Metadata <strong>error</strong> property MUST be <strong>internalError</strong>.
+	<h2>invalidPublicKeyLength</h2>
+	<p>If the byte length of <i>rawPublicKeyBytes</i> does not match the expected public key length for the associated multicodecValue during <a>DID Resolution</a> or <a>DID URL dereferencing</a>,
+		the value of the DID Resolution or DID URL Dereferencing Metadata <strong>error</strong> property MUST be <strong>invalidPublicKeyLength</strong>.
+	<h2>invalidPublicKey</h2>
+	<p>If an invalid public key value is detected during <a>DID Resolution</a> or <a>DID URL dereferencing</a>,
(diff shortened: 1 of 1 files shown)


msporny (approved):

peacekeeper (approved):
Looks useful!

peacekeeper:
The only issue I see is that "invalidDidUrl" is [already defined in DID Core](https://www.w3.org/TR/did-core/#did-url-dereferencing-metadata), so either we don't define it here in DID Resolution, OR we include all DID Core errors also in DID Resolution (and potentially add more details).

aljones15:
@peacekeeper the existing file:///home/liminal18/Programs/nodejs/w3c/did-core/index.html#did-resolution-metadata errors are actually kind of small, so I think we should add them to this section with more specific instructions on when to throw. We should link back to did core so it is known those errors are did core and not specific to did key or did resolution. That is my idea, but consensus could vary. @msporny ideas here?

peacekeeper:
> I think we should add them to this section with more specific instructions on when to throw.

I agree this would be a good idea. Do you want to add that now, or later? Maybe first merge this PR here? (why is it still draft?)

aljones15:
> > I think we should add them to this section with more specific instructions on when to throw.
> 
> I agree this would be a good idea. Do you want to add that now, or later? Maybe first merge this PR here? (why is it still draft?)

Sorry about the delay, but I got massively delayed last week by some more pressing stuff. I will add the did-core stuff today.

peacekeeper (approved):

peacekeeper:
@aljones15 I merged this, and note that I added a few other commits on top of it:
- https://github.com/w3c-ccg/did-resolution/commit/d210f527a407e65712eb5cca40f321993a141265 and https://github.com/w3c-ccg/did-resolution/commit/36f4dd1565427a0f6003d93cbbb68233fa8386a6to improve references to the DID Core sections.
- https://github.com/w3c-ccg/did-resolution/commit/b7079295472a7bfa7cb9419d5ecd5911727c1be5 to re-arrange errors to make DID Core errors show up on top
- https://github.com/w3c-ccg/did-resolution/commit/7b7c2dcdbefb1fdf21b788f8c1fda63185851905 to add representationNotSupported. You're right that DID Resolution currently doesn't talk about it, but it should, and therefore (to be consistent) I think it's best to list all DID Core errors also in DID Resolution.

PR #73, opened 2022-08-10
Add table to map errors to HTTP status codes.

peacekeeper:
This adds a table that maps DID Resolution and DID URL Dereferencing errors to HTTP status codes, which seems easier to read than a list of copy&paste text statements. Also adds an additional status code for the "methodNotSupported" error.

Related to https://github.com/w3c-ccg/did-resolution/issues/72.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1017,7 +1017,7 @@ data-cite="RFC3986#section-2.1">RFC3986 Section 2.1</a>).</li>
 			<li>Dereference the <var>input <a>DID</a></var> or <var>input <a>DID URL</a></var> by executing the
 				<a>DID URL dereferencing</a> algorithm as defined in <a href="#dereferencing"></a></li>
 			<li>If the output of the <a>DID URL dereferencing</a> function contains the <b>dereferencingMetadata</b> property <b>error</b>,
-				then the HTTP response status code MUST be set to the value that corresponds to the value of the of the <b>error</b> property,
+				then the HTTP response status code MUST be set to the value that corresponds to the value of the <b>error</b> property,
 				according to the following table:
 				<table class="simple">
 					<thead>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1016,31 +1016,85 @@ <h2>HTTP(S) Binding</h2>
...
-				<ol class="algorithm">
-					<li>The HTTP response status code MUST be <code>404</code>.</li>
-				</ol>
-			</li>
-			<li>If the output of the <a>DID URL dereferencing</a> function contains the <b>dereferencingMetadata</b> property <b>error</b> with value <b>representationNotSupported</b>:
-				<ol class="algorithm">
-					<li>The HTTP response status code MUST be <code>406</code>.</li>
-				</ol>
+			<li>If the output of the <a>DID URL dereferencing</a> function contains the <b>dereferencingMetadata</b> property <b>error</b>,
+				then the HTTP response status code MUST be set to the value that corresponds to the value of the of the <b>error</b> property,
```
TallTed, line 1020:
```suggestion
				then the HTTP response status code MUST be set to the value that corresponds to the value of the <b>error</b> property,
```

TallTed (approved):
After the minor fix suggested above...

aljones15 (approved):
This makes my life a lot easier, we'll need statusCode for the `did:key` errors too. I guess that will need to either be here or in a new section of the `did:key` spec.

peacekeeper:
> This makes my life a lot easier, we'll need statusCode for the `did:key` errors too. I guess that will need to either be here or in a new section of the did:key spec.

I think the HTTP status code mapping for an error code should be done in the same document where the error itself is defined.

Just as a side note, new errors don't necessarily HAVE to be defined in either the DID Core or DID Resolution spec. If there are errors that are DID method specific, then they can also be defined in the DID method specification. Errors (or any other extension) can also be defined in some other place, e.g. we have already begun to define some errors and other extensions in a DIF repository: https://github.com/decentralized-identity/did-spec-extensions/

So if you want you could move some of the errors out of DID Resolution and into the did:key method spec (either way seems fine to me). OTOH in this specific case, maybe the public key related errors are more generally useful and don't exclusively apply to did:key, not sure.

In any case, no matter where errors or other extension are defined, they can always be listed in the DID Spec Registries.

PR #75, opened 2022-10-01
DIDs are not necessarily under control of the DID subject.

titusz:
see: https://www.w3.org/TR/did-core/#set-2-the-did-subject-is-not-the-did-controller


TallTed (approved):

kdenhartog (approved):
Seems like a reasonable switch

peacekeeper (approved):
Good fix, thanks!

Issue #66, opened 2021-07-15
`versionId`/`versionTime` are not defined for DID resolution


letmaik:
https://w3c-ccg.github.io/did-resolution/#versioning says:
> If a versionId or versionTime DID parameter is provided, the DID resolution algorithm returns a specific version of the DID document.

DID resolution is defined for a DID, not a DID URL, so the only place where such parameters could go is in the DID Resolution Input Metadata, but currently the only property defined there is `accept`.

Am I missing something?

clehner:
I think this specification implies that DID parameters can be used as DID Resolution Input Metadata properties (now called DID Resolution Options in DID Core), in <https://w3c-ccg.github.io/did-resolution/#dereferencing-algorithm>:
> All DID parameters of the input DID URL MUST be passed as input metadata properties to the DID Resolution algorithm.

These properties/options are then passed to the DID method for a Read operation: https://w3c-ccg.github.io/did-resolution/#resolving-algorithm
> 3.1 Besides the input DID, all additional input metadata properties of this algorithm MUST be passed to the Read operation of the input DID method.

Maybe it would help to propose registering `versionId` and `versionTime` in https://www.w3.org/TR/did-spec-registries/ as DID Resolution Input Metadata properties (DID Resolution Options)?

peacekeeper:
Hello, I agree with @clehner 's interpretation.. The intention is that `versionId` and `versionTime` can be passed as resolution options to the resolve() / resolveRepresentation() functions.

And yes the idea was also that if you dereference() a DID URL, then the DID parameters in the DID URL are passed as resolution options to the resolve() / resolveRepresentation() functions.

I just pushed some commits that harmonizes the language in this spec with DID Core (e.g. "input metadata properties" -> "resolution options").

kdenhartog:
> Hello, I agree with @clehner 's interpretation.. The intention is that `versionId` and `versionTime` can be passed as resolution options to the resolve() / resolveRepresentation() functions.
> 
> And yes the idea was also that if you dereference() a DID URL, then the DID parameters in the DID URL are passed as resolution options to the resolve() / resolveRepresentation() functions.

wip-abramson:
I wonder if we can address @clehner's proposal by adding a sentence to the DID Resolution Options - https://w3c.github.io/did-extensions/resolution/#did-resolution-options - that states something like.

`Additionally, all DID URL Parameters are valid DID Resolution Options.`

peacekeeper:
[...]

PR #119, opened 2025-01-31
Add versionId and versionTime to DID resolution options

wip-abramson:
This is to address part of the issue raised in #66 

I have just copied the definitions from DID parameters section - https://w3c.github.io/did-resolution/#did-parameters

Is there a better way to do this. I suppose I could just put some text like:

`See description of the versionId parameter in https://w3c.github.io/did-resolution/#did-parameters`

### DIFF
--- a/index.html
+++ b/index.html
@@ -587,6 +587,29 @@ string">ASCII string</a>. The <a>DID resolver</a> implementation SHOULD use this
 				function and MUST NOT be used with the <code>resolve</code> function.
 			</dd>
 		</dl>
+		<dl>
+			<dt>versionId</dt>
+			<dd>
+				Identifies a specific version of a <a>DID document</a> to be resolved (the
+				version ID could be sequential, or a <a>UUID</a>, or method-specific).
+				If present, the associated value MUST be an <a
+					data-lt="ascii string">ASCII string</a>.
+			</dd>
+		</dl>
+		<dl>
+			<dt>versionTime</dt>
+			<dd>
+				Identifies a certain version timestamp of a <a>DID document</a> to be resolved.
+				That is, the <a>DID document</a> that was valid for a <a>DID</a> at a certain
+				time. If present, the associated value
+				MUST be an <a data-lt="ascii string">ASCII string</a> which is a valid XML
+				datetime value, as defined in section 3.3.7 of <a
+					href="https://www.w3.org/TR/xmlschema11-2/">W3C XML Schema Definition Language
+				(XSD) 1.1 Part 2: Datatypes</a> [[XMLSCHEMA11-2]]. This datetime value MUST be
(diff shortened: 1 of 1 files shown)


ankurdotb (commented):
Yes, this change makes sense. I thought it was already mentioned to behave as such in DID Core.

I'm not sure whether there's clarifying text needed about `versionTime`. For example:

1. DID created on 1st January 2025
2. DID updated on 30th January 2025
3. If `versionTime` passed is:
   1. ..."2025-01-01T00:00:00Z" -> resolves to version of DIDDoc from 1st January
   2. ..."2025-02-06T00:00:00Z" -> resolves to version of DIDDoc from 30th January

i.e., the `versionTime` does not need to match the exact timestamp, and that it means "find the closest **oldest** version at this given date/time".

There's also a question here on what to do if the DID method does _not_ store historical versions. Is this a case worth covering? I suggest in that case:
1. If `versionTime` is equal to or greater than updated date (if updated) or created (if created but never updated) -> return result as normal (basically as if the query param didn't exist)
2. If `versionTime` is older than the above scenario -> return a 404 (this makes more sense in the updated scenario, where the DID method hasn't stored intermediate versions).

TallTed:
> "find the closest oldest version at this given date/time"

I *think* that should be "find the most recent version that is older than this date/time", in other words, "find the version that was in effect at date/time".

index.html
```diff
@@ -587,6 +587,29 @@ <h3>DID Resolution Options</h3>
...
+				If present, the associated value MUST be an <a
+					data-lt="ascii string">ASCII string</a>.
+			</dd>
+		</dl>
+		<dl>
+			<dt>versionTime</dt>
+			<dd>
+				Identifies a certain version timestamp of a <a>DID document</a> to be resolved.
+				That is, the <a>DID document</a> that was valid for a <a>DID</a> at a certain
+				time. If present, the associated value
```
wip-abramson, line 604:
```suggestion
				That is, the most recent version of the <a>DID document</a> that was valid for a <a>DID</a> 
				before the specified `versionTime`. If present, the associated value
```

Something like this @ankurdotb @TallTed?

Note, I will need to update the definition in DID parameters to. But will do that once we have agreed on the change.

ankurdotb: Yes, this makes sense and helps disambiguate

TallTed: Looks good to me

ankurdotb (approved):

peacekeeper (approved):
Looks good to me..

wip-abramson:
Okay, I updated this branch to just reference the definitions of these properties in the DID parameters section.

peacekeeper:
Merging after positive reviews, suggestions incorporated, and discussion in today's WG call.

Issue #90, opened 2024-09-13
More advanced DID URL dereferencing for verificationMethods?


wip-abramson:
As I reread the DID URL dereferencing section I found myself wondering about the focus on service endpoints for the primary resource section. ([Section 4.3.1](https://w3c.github.io/did-resolution/#dereferencing-algorithm-primary))

I guess the simple way to dereference a verificationMethod is through a fragment to the ID of the vm you want. 

e.g. `did:example:1234#keys-1`

But what about a more complex query that returns all verificationMethods with a given verificationRelationship?

I am not exactly sure what that query should look like, but it feels like a useful and common interaction pattern across different DID methods.

Maybe something like `did:example:1234/verificationMethods?rel=assertionMethod`

I think you can also imagine a common dereference like this which includes a fragment to a verificationMethod. Basically, you are checking that the verificationMethod that produced this kind proof was authorized to do so in the DID document.

e.g. for this proof

```
  "proof": {
    "type": "DataIntegrityProof",
    "cryptosuite": "eddsa-rdfc-2022",
    "created": "2021-11-13T18:19:39Z",
    "verificationMethod": "did:example:1234#key-1",
    "proofPurpose": "assertionMethod",
    "proofValue": "z58DAdFfa9SkqZMVPxAQp...jQCrfFPP2oumHKtz"
  }
```

You might want to dereference the following DID URL: `did:example:1234/verificationMethods?rel=assertionMethod#key-1`

I guess per 4.3) anyone could implement this in a custom manner for their DID method. Is it worth trying to bring something like this into the DID resolution spec?

wip-abramson:
I think this is a good idea to add into the DID resolution specification. I expect it to be useful across DID methods as verificationRelationships is a DID core property.

pchampin:
This was discussed during the [did meeting on 10 October 2024](https://www.w3.org/2024/10/10-did-minutes.html#ca33).

<details><summary><i>View the transcript</i></summary>

<h3>DID Resolution Issue/PR Processing</h3>
<p><cite>burn:</cite> Contact the chairs if anyone would suggest an improvement</p>
<p><cite>markus_sabadello:</cite> let's start with new issues</p>
[...]

Issue #87, opened 2024-08-29
Consider use of "Problem Details" in case of errors


peacekeeper:
At the moment, the specification defines various errors that can be returned by the DID Resolution and DID URL Dereferencing functions:
https://w3c.github.io/did-resolution/#errors

During the 29 Aug 2024 DID WG meeting, it has been suggested that the "Problem Details" specification ([RFC9457](https://www.rfc-editor.org/rfc/rfc9457)) could be used here as well.

pchampin:
This was discussed during the [did meeting on 23 January 2025](https://www.w3.org/2025/01/23-did-minutes.html#dcd1).

<details><summary><i>View the transcript</i></summary>

<h4><a href="https://github.com/w3c/did-resolution/issues/87" rel="noopener noreferrer">w3c/<wbr>did-resolution#87</a></h4>
<p><cite>markus_sabadello:</cite> This has been addressed -- exists. We have resolution and dereferencing result. Pending closed.</p>
<p><cite>manu:</cite> Good idea to settle on Problem Details. Handled in other specs and works well. Nice if the DID Resolution spec so there was consistency across the specs (DI, DID, others)/.</p>
<p><cite>ivan:</cite> Formerly -- if have a DID error, inherits the CID error, so that covers it.</p>
<p><cite>Wip:</cite> Argument to put it in the spec.  Any takers to put it in?</p>
<p><cite>manu:</cite> I'll volunteer.</p>

<hr /></details>

wip-abramson:
I am taking a look at addressing this along with #136. Wondering where we landed on:

> But maybe we want to align a bit more than that, i.e. only have a single way of returning (and registering) errors, shared between CID and DID.

If we were to only have a single way following the problem detail approach, I think using the `error` property would make sense. I would be in favor of this approach.

Also, @ankurdotb I think we would follow the CID approach - https://www.w3.org/TR/cid-1.0/#processing-errors - and use the `https://w3id.org/security#` as the base URL.

Finally, as I understand Problem Details, when the Resolver is called over HTTP these errors would be returned separately, or instead of a DID resolution result. Is that correct? 

From the CID spec:

> When exposing these errors through an HTTP interface, implementers SHOULD use [[RFC9457](https://www.w3.org/TR/cid-1.0/#bib-rfc9457)] to encode the error data structure

Or would this kind of response still be a valid application of problem details.

{
  "didDocument": null,
  "didDocumentMetadata": { },
  "didResolutionMetadata": {
[...]

PR #122, opened 2025-02-18
Define verificationRelationship as a dereferencing option

wip-abramson:
This is not complete. Submitting as PR so we can have some discussion.

Some thoughts

1. This PR only adds `verificationRelationship` as a dereferencing option and NOT as query parameter. After reviewing the Retrieve Verification Method algorithm I think this is the way to go - See #90 
2. I expect the algorithm steps need some refining. I attempted to follow the style and pattern currently in use as best I could. I felt this step had to go at the start of the algorithm, because the Retrieve Verification Method algorithm calls the dereferencing algorithm.
3. I wasn't sure about the contentType in the response, I just put application/json???. So we need to update that
4. What about error handling, the Retrieve Verification Method algorithm has a bunch of errors that I expect we want to bubble up to the dereferencingMetadata. Is there any spec text I should put to indicate this though?

***
<a href="https://pr-preview.s3.amazonaws.com/wip-abramson/did-resolution/pull/122.html" title="Last updated on Apr 22, 2025, 3:27 PM UTC (6289d65)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-resolution/122/47a91f5...wip-abramson:6289d65.html" title="Last updated on Apr 22, 2025, 3:27 PM UTC (6289d65)">Diff</a>


index.html
```diff
@@ -1112,6 +1112,22 @@ <h3>DID URL Dereferencing Options</h3>
...
+			<dd>
+				The verificationRelationship that the caller expects the verificationMethod 
+				dereferenced from the DID URL to be authorized for. If present, the 
+				associated value MUST be an <a data-lt="ascii string">ASCII string</a>.  
+				If the DID URL does not dereference to a verificationMethod, or the 
+				verificationMethod is not authorized by the DID document for the specified 
+				verificationRelationship then an error MUST be raised. When present 
+				the DID URL MUST be dereferenced by executing the Retrieve Verification 
+				Method algorithm, see <a data-cite="CID#retrieve-verification-method">Section 3.3</a>
+				of the [[[CID]]] specification.
```
TallTed, line 1128:
```suggestion
				The verificationRelationship for which the caller expects the verificationMethod
				dereferenced from the DID URL to be authorized. If present, the
				associated value MUST be an <a data-lt="ascii string">ASCII string</a>.
				If the DID URL does not dereference to a verificationMethod, or the
				DID document does not authorize the verificationMethod for the specified
				verificationRelationship, then an error MUST be raised. When present,
				the DID URL MUST be dereferenced by executing the Retrieve Verification
				Method algorithm; see <a data-cite="CID#retrieve-verification-method">Section 3.3</a>
				of the [[[CID]]] specification.
```

index.html
```diff
@@ -1172,6 +1188,33 @@ <h2>Algorithm</h2>
 		<section id="dereferencing-algorithm-resource">
 			<h2>Dereferencing the Resource</h2>
 			<ol class="algorithm">
+				<li>If the <var>input dereferencingOptions</var> contains the <code>verificationRelationship</code>
+					[[[#did-url-dereferencing-options]]] property:
+					<ol class="algorithm">
+						<li>Execute the Retrieve Verification Method algorithm, see <a data-cite="CID#retrieve-verification-method">
```
TallTed, line 1194:
```suggestion
						<li>Execute the Retrieve Verification Method algorithm; see <a data-cite="CID#retrieve-verification-method">
```

index.html
```diff
@@ -1172,6 +1188,33 @@ <h2>Algorithm</h2>
...
+							Section 3.3</a>	of the [[[CID]]] specification:
+						<ol class="algorithm">
+							<li>Read the value of the <code>verificationRelationship</code> property of the
+							<var>input dereferencingOptions</var>. This is called the <var>input verificationRelationship</var>.</li>
+							<li>Remove the <code>verificationRelationship</code> property from the <var>input 
+							dereferencingOptions</var></li>
+							<li>Pass the <var>input <a>DID URL</a></var> as the vmIdentifier, the <var>input verificationRelationship</var>
```
TallTed, line 1201:
```suggestion
							<li>Pass the <var>input <a>DID URL</a></var> as the vmIdentifier, the <var>input verificationRelationship</var>,
```

index.html
```diff
@@ -1172,6 +1188,33 @@ <h2>Algorithm</h2>
...
+							<var>input dereferencingOptions</var>. This is called the <var>input verificationRelationship</var>.</li>
+							<li>Remove the <code>verificationRelationship</code> property from the <var>input 
+							dereferencingOptions</var></li>
+							<li>Pass the <var>input <a>DID URL</a></var> as the vmIdentifier, the <var>input verificationRelationship</var> 
+								and the <var>input dereferencingOptions</var> to the <a data-cite="CID#retrieve-verification-method">
+								Retrieve Verification Method</a> algorithm.</li>
+							<li>The result is called the <var>output verificationMethod</var>.</li>
+						</ol>
+						</li>
+						<li>Return the following result:
```
TallTed, line 1207:
```suggestion
						<li>Return the result in the following form:
```

TallTed (changes requested):
Tweaks for clarity

peacekeeper:
I think this is going in the right direction, but I don't think this should be at the start of the algorithm. I think all the "generic" parts of the DID URL Dereferencing algorithm should apply normally. It's true that there is some overlap between our "DID URL Deferencing" algorithm here and the "Retrieve Verification Method" algorithm in CID.

Maybe this should come relatively late in the "DID URL Dereferencing" algorithm and we can just reference a part of the "Retrieve Verification Method" algorithm that's actually specific to processing the verification method, e.g. starting at step 8.

wip-abramson:
Okay, thanks @peacekeeper. I think that makes sense.

I agree I did find it strange to put this at the start. I will take a stab a refactoring in that direction.

wip-abramson:
Okay, how about this?

I think it still needs some refining, but I think this might be a better place to capture the logic. Basically, we dereference the resource first then retrieve the verificationMethod.

There is still some further alignment with the Retrieve Verification Method algorithm required:

- What is the `vmIdentifier`
- What is the `controllerDocumentUrl`

Thinking about the case where: `did:example:1234?versionId=1#vm-2` is dereferenced with `verificationRelationship` option set.

wip-abramson:
Okay, I had another pass at this bringing across steps from the [Retrieve Verification Method](https://w3c.github.io/cid/#retrieve-verification-method algorithm):

- Steps 1-6 are skipped. Handled by the DID resolution algorithm, especially step 3.4 that states the result must be a conformant DID document.
- I brought across 7 and 8
- Step 9 I wasnt sure about. Will this have already been checked?

> If the absolute URL value of verificationMethod.id does not equal vmIdentifier, an error MUST be raised and SHOULD convey an error type of [INVALID_VERIFICATION_METHOD](https://w3c.github.io/cid/#INVALID_VERIFICATION_METHOD).

- Step 10, also not sure. Does the controller property of the verificationMethod have to equal the DID in this case? Can I not include verificationMethods with other DIDs as the controller that are still authorized for a specific verification relationship under my DID document?

> If the absolute URL value of verificationMethod.controller does not equal controllerDocumentUrl, an error MUST be raised and SHOULD convey an error type of [INVALID_VERIFICATION_METHOD](https://w3c.github.io/cid/#INVALID_VERIFICATION_METHOD).

- Steps 11 and 12 brought across.

I think its getting there.

peacekeeper (approved):
I'd be fine with merging this. I think there are still two possible improvements:

1. The following rule should maybe apply every time the dereferencing result is a verification method, not just if the "verificationRelationship" option is present:

> _If verificationMethod is not a conforming verification method ..._

2. We need to align DID Resolution/Dereferencing error reporting with CID error codes. (Also see https://github.com/w3c/did-resolution/issues/87)

But I think we could still merge this PR and address these topics in separate PRs later.

peacekeeper:
In the Echidna workflow logs, I see this message which I guess is blocking the PR:

> Element “ol” not allowed as child of element “ol” in this context.

index.html
```diff
@@ -1301,6 +1317,45 @@ <h2>Dereferencing the Fragment</h2>
 			on the media type ([[RFC2046]]) of the resource, i.e., on the result of
 			<a href="#dereferencing-algorithm-resource"></a>.</p>
 			<ol class="algorithm">
+				<li>If the result of <a href="#dereferencing-algorithm-resource"></a> is an <var>output <a>DID document</a></var>,
+				and the <var>input <a href="#did-url-dereferencing-options">DID URL dereferencing options</a></var> contain the
+				<code>verificationRelationship</code> option:
+				</li>
```
TallTed, line 1323: I think this will fix the echidna error.
```suggestion
```

wip-abramson:
> 2\. We need to align DID Resolution/Dereferencing error reporting with CID error codes. (Also see [Consider use of "Problem Details" in case of errors #87](https://github.com/w3c/did-resolution/issues/87))

I agree, lets tackle this in a future PR.

peacekeeper:
> The other DID URLs that dereference to a verificationMethod are just fragments. E.g. `did:example:1234#key1`
> 
> This just has to dereference to something identified in the DID document as `#key1`. Which could, in theory, be a service endpoint I think.

I agree. This looks good to me now..

peacekeeper:
@wip-abramson looks like merge conflicts need to be resolved here.

wip-abramson:
Okay, think is should be good to go now

peacekeeper:
Merging after approval, and suggestions applied.

Issue #112, opened 2025-01-29
Various issues with links to DID core


wip-abramson:
As I read through the spec I noticed a few related issues to how we are referencing DID core.

The spec references DID core throughout, often specific sections like DID syntax.

e.g. 

```
<a href="https://w3c.github.io/did-core/#did-syntax"></a>
```

This has a number of problems:

1. The URL gives a 404. I should be "https://w3c.github.io/did/...". Or maybe a redirect should be setup?
2. There is no text provided for the link to be placed on. The a tag is empty. There are places in the spec where this renders to things like "See ."
3. Should we in fact be referencing TR space? e.g. "https://www.w3.org/TR/did-core/..". We do this is some places and not others.

Regarding 2, I believe the only place we leave the a tag empty is where we are referencing sections within the spec. e.g. `<a href="#dereferencing"></a>` becomes [5. DID URL Dereferencing](https://w3c.github.io/did-resolution/#dereferencing). Is there some way to get this behaviour for sections in other specs.

Regarding 3, in some places we reference DID Core using [[DID-CORE]], however I think that is when we want to reference the whole spec and not a section of the spec.

I was in the middle of a PR that fixes this, but then realised I could do with some clarification around what the desired fix actually it.

cc @peacekeeper

msporny:
The proper way to do this, IIRC is:

```
See <a data-cite="DID#did-syntax">Section 3.4</a> of the [[[DID]]] specification.

wip-abramson:
Closed as addressed by #124

PR #124, opened 2025-02-20
Minor editorial changes

iherman:
(While reviewing the spec.)

- Copied some CSS for simple tables from the DID core spec. Made the tables more readable (with alternate colors, visually separating the rows)
- The reference to did core should use the respec/specref pattern using data-cite, to be in line with other usages.

### DIFF
--- a/index.html
+++ b/index.html
@@ -91,6 +91,34 @@
 		  display: block;
 		  background-color: #ff9;
 	  }
+
+		table.simple {
+			border-collapse: collapse;
+			margin: 25px 0;
+			min-width: 400px;
+			border: 1px solid #dddddd;
+		}
+		table.simple thead tr {
+			background-color: #005a9c;
+			color: #ffffff;
+			text-align: left;
+		}
+		table.simple th,
+		table.simple td {
+			padding: 12px 15px;
+			vertical-align: top;
+			text-align: left;
+		}
+		table.simple tbody tr {
+			border-bottom: 1px solid #dddddd;
+		}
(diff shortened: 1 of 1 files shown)


peacekeeper (approved):
Thanks, I think this addresses https://github.com/w3c/did-resolution/issues/112 ?

Just wondering if the data-cite should be using `did-core#...` or `DID#...` ?

iherman:
> 
> Just wondering if the data-cite should be using `did-core#...` or `DID#...` ?

yes, did# (or DID#) is an alias to did-core in specref, so it is a possibility, but it also does not really change anything. It is a matter of taste, really.

The question is rather whether we want, at some point, refer to did-1.1 (which is now in specref for a WD) instead of the 1.0 version. That is a WG question, that I did not want to decide on. The question is whether we will ever depend on specifically 1.1 features

(At some point, when DID 1.1 becomes a rec, I would expect did-core (or did) to refer to 1.1.)

wip-abramson:
> Thanks, I think this addresses #112 ?

This does partly address #112, although there are similar issues in the terms.html file. Perhaps @iherman you can catch those too in this PR?

iherman:
> > Thanks, I think this addresses #112 ?
> 
> This does partly address #112, although there are similar issues in the terms.html file. Perhaps @iherman you can catch those too in this PR?

Oops, I did not realize there is also another file. My bad. I have just updated it.

peacekeeper:
Thanks, merging editorial changes.

Issue #85, opened 2024-08-21
Request service by type


philarcher:
I have looked through the issues and can't find one to match this - sincere apologies if I've missed it and this is a duplicate.

In section 4.1.1 there's a note that points to a [5 year old email](https://lists.w3.org/Archives/Public/public-credentials/2019Jun/0028.html) from @dlongley raising the possibility of a DID resolver returning info about a service identified by type rather than by ID. @ChristopherA was not supportive! 

However, we use exactly this kind of request in a [GS1-Conformant resolver](https://ref.gs1.org/standards/resolver/). That is, you can request a particular *type* of associated resource. We use link relation types for this and have a simple parameter of `linkType` so that https://id.gs1.org/01/09506000164908?linkType=gs1:certificationInfo redirects to a resource that, in this case, gives you a load of certificates related to the product identified by the GTIN 9506000164908. https://id.gs1.org/01/09506000164908?linkType=gs1:sustainabilityInfo redirects to the recycling info and so on. (the `gs1` prefix refers to our namespace. The set of [link types we define](https://ref.gs1.org/voc/?show=linktypes) is managed under our standards process).

This is done by leveraging the Linkset, as defined by [RFC9264](https://www.rfc-editor.org/rfc/rfc9264.html) and you can get the whole linkset with `curl -H "Accept: application/linkset+json" https://id.gs1.org/01/09506000164908`.

What happens if you have multiple target resources of the same type? Step forward [HTTP Response code 300](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#redirection_messages) :-) We've implemented this in the [open source version of the resolver](https://github.com/gs1/GS1_DigitalLink_Resolver_CE) but it's not in the production service yet so don't go trying it.

I'm not suggesting a wholesale adoption of this approach in DID Resolution, merely pointing out that requesting a resource by type is useful and has been thought about and engineered in a nearby context. If there is support within the WG for identifying service endpoints by type, the basic approach might be useful.

No one here will care but the use of RFC9264 and the `linkType` parameter to control resolver behavior is included in ISO/IEC 18975, currently at FDIS stage (Final Draft International Standard) which means that, unless something goes horribly wrong, it will be an ISO/IEC standard before the end of the year. Of more importance here is that it's cited multiple times in the [UN Transparency Protocol](https://uncefact.github.io/spec-untp/docs/specification) work being led by Steve Capell. What that work calls an Identity Resolver, or, informally, a link resolver, has this kind of thing in mind.

[...]

Issue #61, opened 2021-02-04
Service Endpoint Construction inconsistency


clehner:
The example (https://w3c-ccg.github.io/did-resolution/#example-11) is inconsistent with the algorithm (https://w3c-ccg.github.io/did-resolution/#algorithm).

The example says:

> Given the following input service endpoint URL:
>
> `https://example.com/messages/8377464`
>
> And given the following input DID URL:
>
> `did:example:123456789abcdefghi?service=messages&relative-ref=%2Fsome%2Fpath%3Fquery#frag`
>
> Then the output service endpoint URL is:
>
> `https://example.com/messages/8377464/some/path?query#frag`

But by following the algorithm, I get this as the output URL:
```
https://example.com/messages/8377464?service=messages&relative-ref=%2Fsome%2Fpath%3Fquery#frag
```
It looks like in the example the `relative-ref` DID parameter is used instead of (or in addition to) the input DID URL path and query.

There is also an example in `did-core` (https://w3c.github.io/did-core/#example-9) for "A resource external to a DID Document", using service and relativeRef:
```
did:example:123?service=agent&relativeRef=/credentials#degree
```

Is using `relativeRef`/`relative-ref` the way to go? If so, how should it interact with the path component and other query parameters in service endpoint construction?

koptan:
I am also facing the same issue, @clehner how did you solve this issue?

clehner:
@koptan
In the [Service Endpoint Construction](https://w3c-ccg.github.io/did-resolution/#service-endpoint-construction) algorithm I use the path and query component from the decoded relativeRef parameter if present rather than from the input DID URL.

https://github.com/spruceid/ssi/blob/80a3f24bc721eba38d150c2912443b9a97cfe71e/src/did_resolve.rs#L549 (2021-02-02)

This should comply with the examples and the definition of relativeRef https://www.w3.org/TR/2022/REC-did-core-20220719/#did-parameters

peacekeeper:
Sorry it took so long to pay attention to this.

> I use the path and query component from the decoded relativeRef parameter if present rather than from the input DID URL.

Yes, this is correct. The current [Service Endpoint Construction](https://w3c.github.io/did-resolution/#service-endpoint-construction) algorithm is wrong and needs to be fixed to use `relativeRef`. That current algorithm in the document is based on a much earlier version of the spec when we still had matrix parameters.

w3cbot:
[...]

Issue #97, opened 2024-11-19
DID `service` parameter does not account for cases where `id` of `service` is a full IRI external to the DID document


trwnh:
Extracted from https://github.com/w3c/did-core/issues/864#issuecomment-2481859443

By my reading of https://w3c.github.io/did-resolution/#dereferencing-algorithm-primary -- it seems like algorithm step 4.1.1 doesn't account for a "full IRI external to the DID document" use case:

> If the input [DID URL](https://w3c.github.io/did-resolution/#dfn-did-urls) contains the [DID parameter](https://www.w3.org/TR/did-core/#did-parameters) `service` and optionally the `relativeRef` DID parameter:
> > [Example 1](https://w3c.github.io/did-resolution/#example-1)
> > did:example:1234?service=files&relativeRef=%2Fmyresume%2Fdoc%3Fversion%3Dlatest
> 
> From the resolved [DID document](https://w3c.github.io/did-resolution/#dfn-did-documents), select the [service endpoint](https://w3c.github.io/did-resolution/#service-endpoint-construction) whose <mark>`id` property contains a fragment which matches the value of the `service` DID parameter</mark> of the input [DID URL](https://w3c.github.io/did-resolution/#dfn-did-urls). This is called the input [service endpoint](https://w3c.github.io/did-resolution/#dfn-service-endpoints).

So for something like

`did:example:abc123?service=https%3A%2F%2Fexternal-node.example`

The resolution algorithm above reads in a way that it would fail to match anything -- certainly there is no fragment matching `service=https%3A%2F%2Fexternal-jsonld-node.example`, and indeed there is no fragment at all!

```json
{
  "@context": "https://www.w3.org/ns/did/v1.1",
  "id": "did:example:abc123",
  "service": {
    "id": "https://external-jsonld-node.example",
    "serviceEndpoint": "https://will-this-ever-match.example"
  }
}
```

In fact, because of the "fragment" matching algo, something like this might be expected to match:

`did:example:abc123?service=example`

```json
{
  "@context": "https://www.w3.org/ns/did/v1.1",
  "id": "did:example:abc123",
  "service": {
    "id": "https://this-part-doesnt-actually-matter.example/#example",
    "serviceEndpoint": "https://i-guess-this-matches.example"
  }
}
```

peacekeeper:
@trwnh Thanks for reporting this, I agree this is a bug that should be fixed.

w3cbot:
[...]

PR #125, opened 2025-02-20
Add serviceType parameter, and update service construction algorithm.

peacekeeper:
This PR addresses two issues:
- Introduces a `serviceType` DID parameter in addition to the existing `service` DID parameter. See https://github.com/w3c/did-resolution/issues/85
- Completely removes the "Service Endpoint Construction" section, and instead uses RFC3986's algorithm on relative reference resolution for processing the `relativeRef` DID parameter. See https://github.com/w3c/did-resolution/issues/61

One aspect I wasn't sure about is what to do if there are multiple services with the same type. At the moment, this PR says that the first matching service will be returned, but we could also return a random matching service, or we could return a list of all matching services and leave it up to the client to pick one. I found an old message by @dlongley on this topic: https://lists.w3.org/Archives/Public/public-credentials/2019Jun/0028.html

### DIFF
--- a/index.html
+++ b/index.html
@@ -251,7 +251,17 @@ did:example:123?service=files&amp;relativeRef=/resume.pdf
 				<code><a>service</a></code>
 			</td>
 			<td>
-				Identifies a service from the <a>DID document</a> by service ID.
+				Identifies a <a data-cite="did-core#services">service</a> from the <a>DID document</a> by service ID.
+				If present, the associated value MUST be an <a
+					data-lt="ascii string">ASCII string</a>.
+			</td>
+		</tr>
+		<tr>
+			<td>
+				<code>serviceType</code>
+			</td>
+			<td>
+				Identifies a <a data-cite="did-core#services">service</a> from the <a>DID document</a> by service type.
 				If present, the associated value MUST be an <a
 					data-lt="ascii string">ASCII string</a>.
 			</td>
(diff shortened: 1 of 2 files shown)


wip-abramson:
I am not sure I agree with you you have implemented this algorithm @peacekeeper. If I am understanding correctly, the algorithm is expecting to always return a single service. Or rather the serviceEndpoint of that single service.

But, if I use the `serviceType` parameter might I not want to return the set of all services in the resolved DID document that are of that type. 

The DID URL: `did:example:1234?serviceType=DIDCommMessaging`

I would expect to return 0 or more services of the `DIDCommMessaging` type in the DID document.

Only if I also included the `service` parameter like: `did:example:1234?serviceType=DIDCommMessaging&service=msg-1`

Would I expect to get a serviceEndpoint for a single service. And only if that service is of the type specified.

index.html
```diff
@@ -251,7 +251,17 @@ <h2>DID Parameters</h2>
 				<code><a>service</a></code>
 			</td>
 			<td>
-				Identifies a service from the <a>DID document</a> by service ID.
+				Identifies a <a data-cite="did-core#services">service</a> from the <a>DID document</a> by service ID.
```
wip-abramson, line 254:
```suggestion
				Identifies a <a data-cite="did-core#services">service</a> from the <a>DID document</a> by service ID and returns the serviceEndpoint for that serivce.
```

index.html
```diff
@@ -251,7 +251,17 @@ <h2>DID Parameters</h2>
...
+				If present, the associated value MUST be an <a
+					data-lt="ascii string">ASCII string</a>.
+			</td>
+		</tr>
+		<tr>
+			<td>
+				<code>serviceType</code>
+			</td>
+			<td>
+				Identifies a <a data-cite="did-core#services">service</a> from the <a>DID document</a> by service type.
```
wip-abramson, line 264:
```suggestion
				Identifies all <a data-cite="did-core#services">services</a> from the <a>DID document</a> by service type and returns a list of serviceEndpoints for these services.
```

wip-abramson: Although I am unsure if this should be the functionality of `serviceType`. Like should it return a list or URLs or a list of `services`?

index.html
```diff
@@ -1229,28 +1239,58 @@ <h2>Dereferencing the Resource</h2>
...
+									property contains a fragment which matches the value of the <code>service</code> DID parameter.</p>
+								<p>If the <var>input <a>DID URL</a></var> contains the
+									<a data-cite="did-core#did-parameters">DID parameter</a> <code>serviceType</code>:
+									Select the <a data-cite="did-core#services">service</a> if its <code>type</code>
+									property matches the value of the <code>serviceType</code> DID parameter.</p>
+							</ul>
+							The selected <a data-cite="did-core#services">services</a> are a list called the <var>resolved <a>services</a></var>.
+						</li>
+						<li>Read the values of the <code>serviceEndpoint</code> properties of the
+							<var>resolved <a>services</a></var>. This list is called the <var>resolved <a>service endpoint</a> URLs</var>.</li>
```
wip-abramson, line 1263: There is an extra complication here that the `serviceEndpoint` property is defined as

> The value of the serviceEndpoint property MUST be a [string](https://infra.spec.whatwg.org/#string), a [map](https://infra.spec.whatwg.org/#string), or a [set](https://infra.spec.whatwg.org/#ordered-set) composed of one or more [strings](https://infra.spec.whatwg.org/#string) and/or [maps](https://infra.spec.whatwg.org/#string). All [string](https://infra.spec.whatwg.org/#string) values MUST be valid [URIs](https://www.w3.org/TR/did-1.0/#dfn-uri) conforming to [[RFC3986](https://www.w3.org/TR/did-1.0/#bib-rfc3986)] and normalized according to the [Normalization and Comparison rules in RFC3986](https://www.rfc-editor.org/rfc/rfc3986#section-6) and to any normalization rules in its applicable [URI](https://www.w3.org/TR/did-1.0/#dfn-uri) scheme specification.
https://www.w3.org/TR/did-1.0/#dfn-serviceendpoint

In the case where a serviceEndpoint is not a string, what should happen?

index.html
peacekeeper: I think I would prefer to keep identification (what does the parameter identify) separate from the dereferencing logic (what gets returned).

Regarding your question, we could support both via the `accept` option.

If `accept` is missing or `accept` == `application/did` or other DID document media type -> return DID document with only the selected services
If `accept` == `text/uri-list` -> return list of service endpoint URLs

What do you think?

wip-abramson: That makes sense to me. It seems this approach could/should also apply to the `service` parameter?

index.html
TallTed: s/serivce/service/

index.html
```diff
@@ -1229,28 +1239,58 @@ <h2>Dereferencing the Resource</h2>
...
-						the <a href="#service-endpoint-construction">Service Endpoint Construction</a> algorithm.</li>
-						<li>The result is called the <var>output <a>service endpoint</a> URL</var>.</li>
-					</ol>
+						<li>From the <var>resolved <a>DID document</a></var>, select all
+							<a data-cite="did-core#services">services</a> which fulfill the following conditions:
+							<ul>
+								<p>If the <var>input <a>DID URL</a></var> contains the
+									<a data-cite="did-core#did-parameters">DID parameter</a> <code>service</code>:
+									Select the <a data-cite="did-core#services">service</a> if its <code>id</code>
+									property contains a fragment which matches the value of the <code>service</code> DID parameter.</p>
```
wip-abramson, line 1254: Wondering if it might also be worth addressing this bug in this PR - https://github.com/w3c/did-resolution/issues/97

I was working on it, then remembered this PR. So I will at least wait until this is in. But maybe we could just address here?

peacekeeper:
@wip-abramson I tried to address some of your comments now, e.g. in https://github.com/w3c/did-resolution/pull/125#discussion_r1965280751 and https://github.com/w3c/did-resolution/pull/125#discussion_r1967656208, and also tried to add some more details of other aspects of how "service", "serviceType", and "relativeRef" are handled. Maybe you can re-review?

index.html
```diff
@@ -1314,7 +1314,11 @@ <h2>Dereferencing the Resource</h2>
 								<li>If the <var>input <a>DID URL</a></var> contains the
 									<a data-cite="did-core#did-parameters">DID parameter</a> <code>service</code>:
 									Select the <a data-cite="did-core#services">service</a> if its <code>id</code>
-									property contains a fragment which matches the value of the <code>service</code> DID parameter.</li>
+									property matches the value of the <code>service</code> DID parameter. In case the <code>id</code>
```
TallTed, line 1317:
```suggestion
									property matches the value of the <code>service</code> DID parameter. If the <code>id</code>
```

index.html
```diff
@@ -1365,16 +1433,16 @@ <h2>Dereferencing the Fragment</h2>
 			on the media type ([[RFC2046]]) of the resource, i.e., on the result of
 			<a href="#dereferencing-algorithm-resource"></a>.</p>
 			<ol class="algorithm">
-				<li>If the result of <a href="#dereferencing-algorithm-resource"></a> is an <var>output <a>service endpoint</a> URL</var>,
+				<li>If the result of <a href="#dereferencing-algorithm-resource"></a> is a <var>selected <a>service endpoint</a> URL</var>,
 				and the <var>input <a>DID URL</a></var> contains a <a>DID fragment</a>:
 				<pre class="example nohighlight">did:example:1234?service=files&relativeRef=%2Fmyresume%2Fdoc%3Fversion%3Dlatest#intro</pre>
 				<ol class="algorithm">
-					<li>If the <var>output <a>service endpoint</a> URL</var>
+					<li>If the <var>select <a>service endpoint</a> URL</var>
```
wip-abramson, line 1440: Wondering if this should be selected throughout this algorithm
```suggestion
					<li>If the <var>selected <a>service endpoint</a> URL</var>
```

index.html
```diff
@@ -1289,36 +1299,94 @@ <h2>Dereferencing the Resource</h2>
...
+										<li>The <strong>relative reference</strong> is the value of the
+											<a data-cite="did-core#did-parameters">DID parameter</a> <code>relativeRef</code>.</li>
+										<li>Update the <var>selected <a>service endpoint</a> URL</var> to
+										the result of the "Reference Resolution" algorithm.</li>
+									</ol>
+								</li>
+							</ol>
+						</li>
+						<li>If the <b>accept</b> <var>input <a href="#did-url-dereferencing-options">DID dereferencing option</a></var>
+							is missing, or if its value is the Media Type of a <a>representation</a> of a <a>DID document</a>:
```
wip-abramson, line 1357: Wondering about the Media Type here. I thought we said there was only one media type for a DID document - `application/did`

index.html
```diff
@@ -279,7 +279,17 @@ <h2>DID Parameters</h2>
...
+				If present, the associated value MUST be an <a
+					data-lt="ascii string">ASCII string</a>.
+			</td>
+		</tr>
+		<tr>
+			<td>
+				<code>serviceType</code>
+			</td>
+			<td>
+				Identifies a <a data-cite="did-core#services">service</a> from the <a>DID document</a> by service type.
```
wip-abramson, line 292:
```suggestion
				Identifies a set of one or more <a data-cite="did-core#services">services</a> from the <a>DID document</a> by service type.
```

index.html
peacekeeper: One that we define in our specs. But my understanding is that anyone can also define different DID document representations with different media types, e.g. CBOR based.

index.html
peacekeeper: I tried to address this in https://github.com/w3c/did-resolution/pull/125/commits/cda94f2b5c9cb12bba2fb476eaa8fcc487ffc8d0, i.e. the value of the service parameter can be an absolute or relative URL.

wip-abramson (approved):

peacekeeper:
Merging, after approval, suggestions applied.

Issue #117, opened 2025-01-31
https://w3c.github.io/did-resolution/contexts/did-resolution-v1.json may not be correct


iherman:
At present, the context file includes

```
didDocument: "https://www.w3.org/ns/did/v1"
```

That is incorrect; the URL is the context file's reference for DID v1, not the URL of the `didDocument` property.

What this shows is that, at some point, the vocabulary for the did resolution must be properly defined, as it was done for the various vocabularies in the VC spec family.

peacekeeper:
I agree, there are some aspects of this context that need to be fixed/improved.

In this case of the `didDocument` property, I think my rationale was that yes, `https://www.w3.org/ns/did/v1` is the URL of the DID v1 context file. But could it not simultaneously ALSO be the URL of the `didDocument` property?

In other words, wouldn't it be somewhat logical that the URL of a property equals the URL of a context file which defines the data model that is the value (range) of the property?

If this makes no sense from an RDF / Linked Data perspective, no problem, happy to change it :)

iherman:
> In this case of the `didDocument` property, I think my rationale was that yes, `https://www.w3.org/ns/did/v1` is the URL of the DID v1 context file. But could it not simultaneously ALSO be the URL of the `didDocument` property?

One could argue that the RDF specification does not say anything, formally, about the Resource that is retrieved by the URI used to identify a resource, so this is not fully incorrect. But I believe it goes against Linked Data practice. When dereferencing a property's URL, one expects an RDF Schema entry definition of a term or, at the minimum, a human readable definition of the term. Not a JSON-LD context file, which is a totally different animal.

(See, for example, the `controller` term, whose identifier is `https://w3id.org/security#controller`, and which dereferences to an HTML+RDFa [definition](https://w3id.org/security#controller) if HTML is requested, or [Turtle equivalent](https://w3id.org/security/vocabulary.ttl#controller) if Turtle is requested.)

I believe the right thing to do is to say:

```
didDocument: "https://w3id.org/did-resolution#didDocument",
```

and, at some point, we should work on the formal vocabularies (including the DID Core vocabulary) to provide a proper definition for the term (and the others). Just repeat what we did for VC. Such vocabulary definition(s) would include a Class definition for `DidDocument`, subclass of `sec:controlledIdentifierDocument`, which would be the range for `didDocument`.

[...]

PR #135, opened 2025-03-15
Fix definition of "didDocument" in JSON-LD context

peacekeeper:
Addresses https://github.com/w3c/did-resolution/issues/117

### DIFF
--- a/contexts/did-resolution-v1.json
+++ b/contexts/did-resolution-v1.json
@@ -6,7 +6,7 @@
     "type": "@type",
     "xsd": "http://www.w3.org/2001/XMLSchema#",
 
-    "didDocument": "https://www.w3.org/ns/did/v1",
+    "didDocument": "https://w3id.org/did-resolution#didDocument",
 
     "didResolutionMetadata": "https://w3id.org/did-resolution#didResolutionMetadata",
     "contentType": "https://w3id.org/did-resolution#contentType",


iherman (approved):
I am fine with the change for consistency for now. 

However, at some point in time, though, we will have to discuss whether (a) we need JSON-LD for the did resolution terms in the first place and (b) if the answer is yes, do we need a separate namespace or should we simply add it to the did vocabulary namespace. Let us remind ourselves that the latter has only 1-2 terms (related to services); having separate namespaces for 1-2 terms looks like an administrative overkill to me.

peacekeeper:
Thanks @iherman . I don't have a strong opinion here, yes maybe JSON (without LD) is sufficient.

One complication I see is that this would also change the media type. So far, we have been using these two media types for a DID Resolution Result and a DID URL Dereferencing Result:

- `application/ld+json;profile="https://w3id.org/did-resolution"`
- `application/ld+json;profile="https://w3id.org/did-url-dereferencing"`

> Let us remind ourselves that the latter has only 1-2 terms (related to services)

I don't understand what you mean by this... The current [DID v1.1rc1 context](https://github.com/w3c/did/blob/main/contexts/v1.1rc1) has more than 1-2 terms, and seems equivalent to the [CID 1.0 context](https://github.com/w3c/cid/blob/main/context/v1.jsonld).

iherman:
> > Let us remind ourselves that the latter has only 1-2 terms (related to services)
> 
> I don't understand what you mean by this... The current [DID v1.1rc1 context](https://github.com/w3c/did/blob/main/contexts/v1.1rc1) has more than 1-2 terms, and seems equivalent to the [CID 1.0 context](https://github.com/w3c/cid/blob/main/context/v1.jsonld).

I am talking about the vocabulary, ie, the terms defined by a specific document, not the JSON-LD Context. A context file does not _define_ anything, it is just a glorified transformer from a string to a URL :-). If you look at the context file, most of the terms are in the [security vocabulary](https://w3c.github.io/vc-data-integrity/vocab/security/vocabulary.html), except for `service` and the `serviceEndpoint`.

peacekeeper:
As discussed today, let's create a separate issue for the "JSON-LD or not" question and related topics such as the vocabulary.

dlongley (approved):

peacekeeper:
Merging after multiple approvals.

Issue #53, opened 2020-02-03
Disallow JSON-LD remote context retrieval


peacekeeper:
@awoie made a good suggestion that JSON-LD context files must not be retrieved from a remote location in a production environment. The DID Resolution spec could state this explicitly. See https://lists.w3.org/Archives/Public/public-credentials/2020Jan/0066.html

awoie:
A lot of people will probably only implement normative statements from the spec, so I think we should make that normative. After the conversation on the mailing list, I have the perception that nobody uses remote retrieval in practice and it is considered to be bad practice. So, we should make this a normative statement.

w3cbot:
This was discussed during the [#did meeting on 30 January 2025](https://www.w3.org/2025/01/30-did-minutes.html#5a42).

<details><summary><i>View the transcript</i></summary>

<h4><a href="https://github.com/w3c/did-resolution/issues/53" rel="noopener noreferrer">w3c/<wbr>did-resolution#53</a></h4>
<p><cite>Wip:</cite> issue 53 - disallow remote context retrieval<br>
<span>… this is suggesting a normative statement</span><br>
<span>… this is about nor making remote requests for contexts in a production environment</span></p>
<p><cite>markus_sabadello:</cite> this often comes up with any JSON-LD context, that they should be cached to avoid man-in-the-middle attacks<br>
<span>… there is information in the VC specification that talks about needing to cache contexts</span><br>
<span>… so we could add something similar here</span><br>
<span>… it could potentially go in the security considerations section</span></p>
<p><cite>Wip:</cite> something like production DID resolvers MUST cache contexts?<br>
<span>… pretty straightforward, and security considerations seems like a reasonable location</span><br>
<span>… any takers?</span><br>
<span>… I'll take it</span></p>
<p><cite>Wip:</cite> I wanted to open the door to JoeAndrieu and smccown</p>

<hr /></details>

w3cbot:
This was discussed during the [#did meeting on 08 May 2025](https://www.w3.org/2025/05/08-did-minutes.html#7d5e).

<details><summary><i>View the transcript</i></summary>

<h3>DID Resolution Issue Processing</h3>
<p><a href="https://github.com/w3c/did-resolution/issues/53" rel="noopener noreferrer">w3c/<wbr>did-resolution#53</a></p>
<p>Disallow JSON-LD remote context retrieval&nbsp;#53EditNew issue</p>
[...]

PR #147, opened 2025-05-08
Add security consideration about JSON-LD Context Integrity

ottomorac:
Add security consideration note about JSON-LD Context Integrity in order to address https://github.com/w3c/did-resolution/issues/53 .

I placed it under a new security considerations item called "JSON-LD Context Integrity".

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-resolution/pull/147.html" title="Last updated on Jun 5, 2025, 9:08 PM UTC (e7c34c3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-resolution/147/1508359...e7c34c3.html" title="Last updated on Jun 5, 2025, 9:08 PM UTC (e7c34c3)">Diff</a>


index.html
```diff
@@ -2345,23 +2345,27 @@ <h2>Authentication/Authorization</h2>
...
+					</ul>
+				<p>Caching behavior can be controlled by configuration of the <a>DID resolver</a>,
+					by the <code>noCache</code> resolution option, or by contents of the DID document
+					(e.g., a `cacheMaxTtl` field), or by a combination of these properties.</p>
+				<p class="issue" data-number="10">See corresponding open issue.</p>
+				<p class="issue">Perhaps we can re-use caching mechanisms of other protocols such as HTTP.</p>
+			<h3>JSON-LD Context Integrity</h3>
+			<p>If the contexts are fetched live each time, an attacker could alter the context file (by compromising the server 
+				or intercepting the request via a man in the middle attack). Therefore a <a>DID resolver</a> that uses JSON-LD 
+				MUST cache contexts.</p>
```
wip-abramson, line 2368:
```suggestion
			<p>If <a href="https://www.w3.org/TR/json-ld11/#the-context">JSON-LD Context</a> files are fetched from a remote location each time, an attacker could alter the context file (by compromising the server 
				or intercepting the request via a man in the middle attack). Therefore a production <a>DID resolver</a> that uses JSON-LD 
				MUST NOT retrieve context files from remote locations.</p>
```

Maybe something like this. I think it isn't even so much about caching, which would still allow contexts to be fetched from a remote service once. Production DID resolvers should have local access to the contexts they require.

index.html
```diff
@@ -2345,23 +2345,27 @@ <h2>Authentication/Authorization</h2>
...
-			</ul>
-		<p>Caching behavior can be controlled by configuration of the <a>DID resolver</a>,
-			by the <code>noCache</code> resolution option, or by contents of the DID document
-			(e.g., a `cacheMaxTtl` field), or by a combination of these properties.</p>
-		<p class="issue" data-number="10">See corresponding open issue.</p>
-		<p class="issue">Perhaps we can re-use caching mechanisms of other protocols such as HTTP.</p>
-	
+			<h3>Caching of DID documents</h3>
+				<p>A <a>DID resolver</a> may maintain a generic cache of <a>DID documents</a>. It may also
+					maintain caches specific to certain <a>DID methods</a>.</p>
```
msporny, line 2350: We need to make sure we use RFC language here -- so, uppercase MAY, SHOULD, MUST.

These statements are probably SHOULD statements -- because caching is a best practice.

index.html
```diff
@@ -2345,23 +2345,27 @@ <h2>Authentication/Authorization</h2>
...
+				<p>The <code>noCache</code> resolution option can be used to
```
msporny, line 2351: We should warn that turning off the cache is also an attack vector... so resolvers might just return an error if they detect something amiss.

wip-abramson:
Changes look good @ottomorac. Wondering if want to bring any additional spec text across from the VCDM spec - https://w3c.github.io/vc-data-model/#base-context.

Specifically, identifying the hash value of the DID context seems useful. This could be done in a separate PR when we have finalised the context.

cc @msporny

dlongley:
Since this topic comes up in a number of places I figured I'd write a bit more here:

The general security requirement here is that any application, such as a DID resolver, must use a trusted copy of the context(s) that its logic is written against. One can think of these contexts as being known at "compile time" -- as opposed to contexts that the application may encounter at "runtime".

As an analogy, this is not too different from ensuring your application does not load some code dependency (a software library/module) over the internet from an untrusted location or without some appropriate integrity check. Calling into a library without these protections is unsafe, just like consuming a document with properties that are only *supposedly* defined a way you expect is unsafe.

Of course, this is why document-based validation, e.g., JSON schema checks in the JSON ecosystem, are typically performed on incoming documents. This allows your application to be sure that a document is "shaped" in the way the its subsequent logic expects (e.g., the document has the property "foo" with a string value in it).

JSON-LD documents carry additional semantic information, allowing an application to *also* ensure that, for example, the property "foo" has an intended global meaning, not just a "shape" that happens to be compatible with your application. This helps to further reduce confusion and provide more confidence that the subsequent logic in your application will execute properly.

While it may seem obvious, it is important to note that in both cases the logic of your application was written prior to receiving any document. The logic of your application natively understands (expects) a particular document shape or "type" and particular semantics for its properties. This means this was baked in / written into the application when it was created -- so both the expected JSON schema and the JSON-LD context(s) were necessarily known at that time and the logic of your application is *bound* to them. One can think of this as an interface your application has. To use your application the inputs must match these constraints.

A reasonable expectation is then that your application only load its JSON schema and its natively understood JSON-LD contexts in a trusted way. The simplest way to do that is for your application to have local copies of these that are used when resolving them; it is common to publish applications with embedded JSON schemas and the same can be done with the JSON-LD contexts that the application is also bound to. This handles the "compile time" requirements.

So, as for how your application handles other contexts that it only later encounters at runtime? You have a choice: it can either reject them outright, requiring others to make the inputs they feed you conform to your application's interface -- or you can accept arbitrary documents (that have arbitrary remote contexts and potentially even arbitrary shapes) and then use the JSON-LD API to transform the documents into the contexts and shapes that your application accepts. As always, some transformations might fail to produce anything acceptable anyway.

For interoperability purposes, the former is much simpler: Just tell data producers that they must use a particular context and document shape if they want their inputs to be commonly accepted by consumers. If someone wants to write an application that can consume many different types of documents (perhaps one that combines or does something else interesting with documents from many different ecosystems) they are of course welcome to do so and an interesting innovation may arise. But everyone can tell what global semantics were intended by authors who produce documents that are self-describing, e.g., documents that carry "context" information, even if those semantics aren't understood (and perhaps are rejected) by a particular consumer.

Finally, but still importantly, even if you loaded an arbitrary JSON schema (or arbitrary JSON-LD context) from the internet and its contents matched some content-integrity hash, your application logic wasn't previously written with the knowledge of this content. So it is not safe to directly consume a document that matched this schema based on that fact alone. It still needs to be transformed into a shape and context that your application's logic was written against. This means that a content-integrity hash on an unknown JSON schema or JSON-LD context does little for you. The hash only matters here *when it is already familiar to you*. Note that this is not specific to JSON schema, JSON-LD contexts, or any particular format but rather a general principle. Most people would not be happy to shout any random message they received into the public square just because its contents matched a hash.

*Every* message has a corresponding hash.

ottomorac:
Thanks @dlongley and @peacekeeper for your comments. 

I added the following verbiage to end of my security consideration:

> Instead a DID resolver SHOULD use a trusted copy of the context file (for instance a local copy of the JSON-LD file obtained prior to resolution).

I opted to use "SHOULD" because if we used "MUST" then that would require precisely defining what a "trusted copy" is. I personally struggle with that a bit, as it seems there are several ways of solving that (embedded contexts, obtaining a copy of the JSON-LD ahead of time, etc.). Let me know if you disagree with this approach or have other wording suggestions.

index.html
```diff
@@ -2361,7 +2361,11 @@ <h2>Caching</h2>
 			(e.g., a `cacheMaxTtl` field), or by a combination of these properties.</p>
 		<p class="issue" data-number="10">See corresponding open issue.</p>
 		<p class="issue">Perhaps we can re-use caching mechanisms of other protocols such as HTTP.</p>
-	
+	</section>
+	<section id="json-ld-integrity">
+		<h2>JSON-LD Context Integrity</h2>
+			<p>If <a href="https://www.w3.org/TR/json-ld11/#the-context">JSON-LD Context</a> files are fetched from a remote location each time, an attacker could alter the context file (by compromising the server 
+				or intercepting the request via a man in the middle attack). Therefore a production <a>DID resolver</a> that uses JSON-LD MUST NOT retrieve context files from remote locations. Instead a <a>DID resolver</a> SHOULD use a trusted copy of the context file (for instance a local copy of the JSON-LD file obtained prior to resolution).</p>
```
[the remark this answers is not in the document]
ottomorac: Thanks Dan. Agree with your suggestions.

index.html
```diff
@@ -2361,7 +2361,11 @@ <h2>Caching</h2>
 			(e.g., a `cacheMaxTtl` field), or by a combination of these properties.</p>
 		<p class="issue" data-number="10">See corresponding open issue.</p>
 		<p class="issue">Perhaps we can re-use caching mechanisms of other protocols such as HTTP.</p>
-	
+	</section>
+	<section id="json-ld-integrity">
+		<h2>JSON-LD Context Integrity</h2>
+			<p>If <a href="https://www.w3.org/TR/json-ld11/#the-context">JSON-LD Context</a> files are fetched from a remote location each time, an attacker could alter the context file (by compromising the server
```
peacekeeper, line 2367:
```suggestion
			<p>If <a href="https://www.w3.org/TR/json-ld11/#the-context">JSON-LD Context</a> files are fetched from a remote location, an attacker could alter the context file (by compromising the server 
```

peacekeeper (approved):
I just have one small suggestion to remove the "each time" wording. Otherwise this looks good to me. But we should wait for the JSON-LD experts on this thread to also review the PR.

index.html
```diff
@@ -2361,7 +2361,11 @@ <h2>Caching</h2>
 			(e.g., a `cacheMaxTtl` field), or by a combination of these properties.</p>
 		<p class="issue" data-number="10">See corresponding open issue.</p>
 		<p class="issue">Perhaps we can re-use caching mechanisms of other protocols such as HTTP.</p>
-	
+	</section>
+	<section id="json-ld-integrity">
+		<h2>JSON-LD Context Integrity</h2>
+			<p>If <a href="https://www.w3.org/TR/json-ld11/#the-context">JSON-LD Context</a> files are fetched from a remote location, an attacker could alter the context file (by compromising the server 
+				or intercepting the request via a man-in-the-middle attack). Therefore, a production <a>DID resolver</a> that uses JSON-LD MUST NOT retrieve context files from remote locations. Instead, a <a>DID resolver</a> SHOULD use a trusted copy of the context file (for instance, a local copy of the context file obtained prior to resolution).</p>
```
dlongley, line 2368: Certainly the text that is currently here in this PR is the strongest possible security advice and it will suffice from a security perspective.

However, DID resolvers can still be secure even if it is relaxed a little, but really only more knowledgeable developers should venture there. I'm not sure if we want to add some additional language to offer flexibility for writing more advanced DID resolvers -- so as to not unduly harm innovation / extensibility.

The additional technical nuance here is:

JSON-LD contexts can be retrieved from remote locations if the JSON-LD API is used to subsequently compact any document that uses them to context(s) that are *not* retrieved from remote locations.

So, a DID resolver must either only strictly accept contexts it can load locally, or it must be equipped with a JSON-LD library that will "compact away" any contexts it loads remotely.

ottomorac: Thanks Dave for this additional suggestion. We discussed in the WG that restricting remote context retrieval was indeed going a bit far. In agreement with the group today, I have added the verbiage used in the VC Data Model around hashing of JSON-LD contexts with some adjustments so that it is written from the perspective of a did resolver. Hope this fits better in addressing the security concern whilst not restricting remote retrieval of context files.

ottomorac:
Hi All, Thanks for your suggestions. After the wg discussion feedback from today; I have added the verbiage used in the VC Data Model around hashing of JSON-LD contexts with some adjustments so that it is written from the perspective of a DID resolver. Hope this works better.

wip-abramson (approved):

peacekeeper (approved):

Issue #43, opened 2019-09-04
Handling DID resolution after a fork


kdenhartog:
@jandrieu raised an interesting question at RWoT that we want to consider in the resolution specification.

Let's assume this scenario:

Alice has `DID:btcr:12345` that exists pre-fork.
Later on the network which `btcr` operates on forks into two separate forks.

How should the resolver dereference the "definitive" DID Document?

Should we take into account a `fork` matrix parameter or modify the identifier to account for the fork such as `did:btcr:12345` -> `did:btcr:fork1:12345`?

brentzundel:
My understanding is that the DID method specifies the network to be used
for resolution. So if the network forks, it would be up to the DID method
to specify which fork holds the "one true DID doc"

peacekeeper:
> Should we take into account a `fork` matrix parameter or modify the identifier to account for the fork such as `did:btcr:12345` -> `did:btcr:fork1:12345`?

I think modifying the identifier would most likely be a bad idea. If we use a matrix parameter, then it should probably be method-specific. Its value should somehow indicate which fork to work on, e.g. by providing a block hash that is expected at a certain block height.

Maybe something like `did:btcr:xz35-jznz-q6mr-7q6;btcr:block-123456=ae676e00`.

Of course if that parameter isn't present, then as @brentzundel says the method would be responsible to somehow figure out what is the "true DID doc".

burnburn:
Note that at an earlier IIW forking came up as one of many different potential DID infrastructure failures that need to be considered.  See https://github.com/windley/IIW_homepage/raw/gh-pages/assets/proceedings/IIWXXVIII_Book_of_Proceedings_2019A.pdf  page 88

wip-abramson:
I am not sure this is something that the resolution spec needs to address.

To me, DID Methods specify the VDR / VDRs that they operate over. 

In the case of btcr that is the Bitcoin blockchain (or its testnets), with the nodes operating the network determining the valid set of blocks. While forks are possible, they are unlikely to be long lived. The fork with the most hash power will be able to produce the most blocks and will considered the Bitcoin blockchain.

In the case of the Bitcoin blockchain forking into a different network, as it has done many times (e.g. https://www.bitcoinsv.com/), then this should be considered a new VDR and not part of the btcr DID method. 

[...]

PR #155, opened 2025-05-28
Ottomorac add vdr forks consideration

ottomorac:
Added security consideration for VDR forks in order to address https://github.com/w3c/did-resolution/issues/43

### DIFF
--- a/index.html
+++ b/index.html
@@ -2412,7 +2412,17 @@ curl -X GET https://resolver.example/1.0/identifiers/did:sov:WRfXPg8dantKVubE3HX
 			<p class="issue" data-number="12">See corresponding open issue.</p>
 
 	</section>
-	
+
+	<section id="vdr-forks">
+		<h2>VDR Network Forks</h2>
+		<p>DID Methods that utilize VDRs of distributed systems as (such as distributed ledgers or other distributed systems) need to manage the potential for network forks taking place. A <a>DID method</a> that uses a distributed system as a VDR SHOULD disambiguate the VDR they are using in the case of such forks by either:</p>
+		<ul>
+			<li>Referencing a well-known public source or resolution library that can distinguish between forked networks in the did method specification</li>
+			<li>Maintaining a manual mapping in the did method specification of the method-specific-id subdivisions to the various VDRs</li>
+			<li>Adding a new did parameter to specify which VDR is being used when network forks occur</li>
+		</ul>
+		<p>In addition a <a>DID Resolver</a> that is aware of any VDR network forks should warn users that these forks may affect resolution results.</p>
+	</section>
 	<section id="non-did-identifiers">
 		<h2>Non-DID Identifiers</h2>
 		<p class="issue" data-number="8">There is discussion on the relationship between <a>DID resolution</a> and
--- a/index.html
+++ b/index.html
@@ -2421,7 +2421,7 @@ curl -X GET https://resolver.example/1.0/identifiers/did:sov:WRfXPg8dantKVubE3HX
 			<li>Maintaining a manual mapping in the did method specification of the method-specific-id subdivisions to the various VDRs</li>
 			<li>Adding a new did parameter to specify which VDR is being used when network forks occur</li>
 		</ul>
-		<p>In addition a <a>DID Resolver</a> that is aware of any VDR network forks should warn users that these forks may affect resolution results.</p>
+		<p>In addition a <a>DID Resolver</a> that is aware of any VDR network forks SHOULD warn users that these forks may affect resolution results.</p>
 	</section>
 	<section id="non-did-identifiers">
 		<h2>Non-DID Identifiers</h2>


index.html
```diff
@@ -2412,7 +2412,17 @@ <h2>Versioning</h2>
 			<p class="issue" data-number="12">See corresponding open issue.</p>
 
 	</section>
-	
+
+	<section id="vdr-forks">
+		<h2>VDR Network Forks</h2>
+		<p>DID Methods that utilize VDRs of distributed systems (such as distributed ledgers or other distributed systems) need to manage the potential for network forks taking place. A <a>DID method</a> that uses a distributed system as a <a>verifiable data registry</a> SHOULD disambiguate the VDR they are using in the case of such forks by either:</p>
```
peacekeeper, line 2418:
```suggestion
		<p><a>DID methods</a> that utilize VDRs of distributed systems (such as distributed ledgers or other distributed systems) need to manage the potential for network forks taking place. A <a>DID method</a> that uses a distributed system as a <a>verifiable data registry</a> SHOULD disambiguate the VDR they are using in the case of such forks by either:</p>
```

index.html
```diff
@@ -2412,7 +2412,17 @@ <h2>Versioning</h2>
...
+		<ul>
+			<li>Referencing a well-known public source or resolution library that can distinguish between forked networks in the did method specification</li>
```
peacekeeper, line 2420:
```suggestion
			<li>Referencing a well-known public source or resolution library that can distinguish between forked networks in the DID method specification</li>
```

index.html
```diff
@@ -2412,7 +2412,17 @@ <h2>Versioning</h2>
...
+			<li>Maintaining a manual mapping in the did method specification of the method-specific-id subdivisions to the various VDRs</li>
```
peacekeeper, line 2421:
```suggestion
			<li>Maintaining a manual mapping in the DID method specification of the <code>method-specific-id</code> subdivisions to the various VDRs</li>
```

index.html
```diff
@@ -2412,7 +2412,17 @@ <h2>Versioning</h2>
...
+			<li>Adding a new did parameter to specify which VDR is being used when network forks occur</li>
```
peacekeeper, line 2422:
```suggestion
			<li>Adding a new DID parameter to specify which VDR is being used when network forks occur</li>
```

index.html
```diff
@@ -2412,7 +2412,17 @@ <h2>Versioning</h2>
...
+		</ul>
+		<p>In addition a <a>DID Resolver</a> that is aware of any VDR network forks SHOULD warn users that these forks may affect resolution results.</p>
```
peacekeeper, line 2424:
```suggestion
		<p>In addition a <a>DID resolver</a> that is aware of any VDR network forks SHOULD warn users that these forks may affect resolution results.</p>
```

peacekeeper (approved):
Here are some small editorial suggestions, but other than that this makes sense to me!

index.html
wip-abramson: I think I agree with this, but does that mean this spec should be defining such a warning?

index.html
wip-abramson: I wonder if this SHOULD statement is better here - https://www.w3.org/TR/did/upcoming/#methods
Although, since we can only do class 3 changes perhaps that is not possible in this iteration.

ottomorac:
Update after wg discussion on 29-may to remove the guidance and indicate that did methods should manage the disambiguation of the VDR forks in the corresponding did method spec.

wip-abramson:
This looks good, although I wonder about the SHOULD.

To me the DID Resolution spec should be placing requirements on implementations of DID Resolvers and Dereferencers. Not DID Methods. 

Also, could you split the line to be ~ < 80 chars

index.html
```diff
@@ -2462,7 +2462,11 @@ <h2>Versioning</h2>
 			<p class="issue" data-number="12">See corresponding open issue.</p>
 
 	</section>
-	
+
+	<section id="vdr-forks">
+		<h2>VDR Network Forks</h2>
+		<p><a>DID methods</a> that utilize VDRs of distributed systems (such as distributed ledgers or other distributed systems) need to manage the potential for network forks taking place. A <a>DID method</a> that uses a distributed system as a <a>verifiable data registry</a> SHOULD disambiguate the VDR they are using in the case of such forks by specifying a method to do so in the corresponding <a>DID method</a> specification.</p>
```
TallTed, line 2468:
```suggestion
		<p><a>DID methods</a> that use a distributed system (such as a distributed ledger or other distributed system) as a VDR (<a>verifiable data registry</a>) need to manage the potential that network forks may occur. Therefore, the specification of a <a>DID method</a> that uses a distributed system as a VDR SHOULD specify a means by which the VDR they are using can be disambiguated from such forks.</p>
```

wip-abramson (approved):
Minor suggestion, but either way LGTM

PR #205, opened 2025-09-25
Remove auto-publication of pull requests.

msporny:
We have been publishing every pull request to W3C TR space as a Working Draft. This is not the correct procedure for PRs. This PR fixes our auto-publication to only kick in when something is merged to the gh-pages branch (which is our main branch, which we should also fix to be `main` and not `gh-pages` since we no longer use the legacy gh-pages publication mechanism).


pchampin:
I would suggest to comment out the lines rather than removing them (easier to revert or reverse engineer).

pchampin (approved):

peacekeeper:
Thanks for the fix, merging.

PR #215, opened 2025-10-09
Standardize error responses per id (#212)

### LINKED ISSUE #212: Standardize error responses with RFC9457ProblemDetails
Hi,
While reviewing the OpenAPI spec, I noticed the error responses (e.g., `404`, `501`) only have a `description` and lack a `content` schema. This leaves the structure of the error body undefined for client implementers.

The spec already includes the `RFC9457ProblemDetails` schema, which is perfect for this purpose. I propose updating all `4xx` and `5xx` error responses to reference this schema.

For example, the `404` response could be:

```yaml
"404":
  description: DID or DID URL not found.
  content:
    application/problem+json:
      schema:
        $ref: "#/components/schemas/RFC9457ProblemDetails"
```

This change would make the API's error handling consistent and more predictable.

Thanks.

erdemonal:
Fixes #212

This pull request updates the OpenAPI specification to provide a consistent error response structure, as discussed in the issue.

It adds a `content` block referencing `RFC9457ProblemDetails` to all `4xx` and `5xx` error responses.

### DIFF
--- a/openapi/openapi.yaml
+++ b/openapi/openapi.yaml
@@ -150,21 +150,30 @@ paths:
         "400":
           description: Invalid DID or DID URL.
           content:
-            application/problem+json:
+            application/did-resolution:
               schema:
-                $ref: "#/components/schemas/RFC9457ProblemDetails"
+                $ref: "#/components/schemas/ResolutionResult"
+            application/did-url-dereferencing:
+              schema:
+                $ref: "#/components/schemas/DereferencingResult"
         "404":
           description: DID or DID URL not found.
           content:
-            application/problem+json:
+            application/did-resolution:
               schema:
-                $ref: "#/components/schemas/RFC9457ProblemDetails"
+                $ref: "#/components/schemas/ResolutionResult"
+            application/did-url-dereferencing:
+              schema:
+                $ref: "#/components/schemas/DereferencingResult"
         "406":
           description: Representation not supported.
           content:
-            application/problem+json:
+            application/did-resolution:
               schema:
-                $ref: "#/components/schemas/RFC9457ProblemDetails"
+                $ref: "#/components/schemas/ResolutionResult"
+            application/did-url-dereferencing:
+              schema:
+                $ref: "#/components/schemas/DereferencingResult"
         "410":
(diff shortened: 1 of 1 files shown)


peacekeeper:
@erdemonal11 I know I agreed with your idea in my comment https://github.com/w3c/did-resolution/issues/212#issuecomment-3386218963, but now after reviewing the PR, I am wondering if this change is really the right thing to do.

My interpretation of the current spec text is that in case of an error, a resolver would still return a full `ResolutionResult` (which can contain a `RFC9457ProblemDetails`), rather than returning only the `RFC9457ProblemDetails`.

But maybe we should discuss this in the WG.

In other words, I think the question is whether the error response content should look like this:

```json
{
    "didDocument": null,
    "didResolutionMetadata": {
        "error": {
            "type": "NOT_FOUND",
            "title": "The DID or DID document was not found.",
            "detail": "No resolve result for did:method:BDrEcHc8Tb4Lb2VyQZWEDE"
        }
    },
    "didDocumentMetadata": {}
}
```

Or like this:

```json
{
    "type": "NOT_FOUND",
    "title": "The DID or DID document was not found.",
    "detail": "No resolve result for did:method:BDrEcHc8Tb4Lb2VyQZWEDE"
}
```

peacekeeper:
@erdemonal11 We discussed this in the DID WG meeting yesterday:
https://www.w3.org/2025/10/16-did-minutes.html#d627

In this meeting, we thought that it's better to leave things as-is. In other words, in case of an error, we include the `RFC9457ProblemDetails` as part of the DID resolution metadata, rather than returning only the `RFC9457ProblemDetails` itself.

Would this be fine with you, or do you see any problems with this?

erdemonal11:
@peacekeeper It is fine for me I checked and I don't see any problems.

peacekeeper:
> @peacekeeper It is fine for me I checked and I don't see any problems.

Thank you, in this case should we close this PR here?

Alternatively you could update the PR so the 4xx and 5xx responses reference `ResolutionResult` or `DereferencingResult`, instead of `RFC9457ProblemDetails`.

Issue #201, opened 2025-09-22
`relativeRef` should prefer UTF-8 for percent encoding


aphillips:
DID Parameters
https://www.w3.org/TR/did-resolution/#did-parameters

> A relative [URI](https://www.w3.org/TR/did-resolution/#dfn-uri) reference according to [RFC3986 Section 4.2](https://www.rfc-editor.org/rfc/rfc3986#section-4.2) that identifies a [resource](https://www.w3.org/TR/did-resolution/#dfn-resources) at a [service endpoint](https://www.w3.org/TR/did-resolution/#dfn-service-endpoints), which is selected from a [DID document](https://www.w3.org/TR/did-resolution/#dfn-did-documents) by using the service parameter. If present, the associated value MUST be an [ASCII string](https://infra.spec.whatwg.org/#ascii-string) and MUST use percent-encoding for certain characters as specified in [RFC3986 Section 2.1](https://www.rfc-editor.org/rfc/rfc3986#section-2.1).

The above requirement doesn't specify a specific character encoding to be used when percent-encoding URI-disallowed characters. Specifying that UTF-8 must be used would help interoperability. If the goal is to allow **_any_** octets to be transmitted, this should be made clear. 

Note that the text does not say what the "certain characters" are. 3986 2.1 only specifies the percent-encoding mechanism, not what the characters are. A casual reader might assume that only **ASCII** characters are permitted, with the percent encoding reserved for problem characters like `=` that are also ASCII.

> [!NOTE]
> An example later in the spec shows a relative reference that includes text that non-English users might reasonably
> expect to include non-ASCII data.
> https://www.w3.org/TR/did-resolution/#example-5
> > did:example:1234?service=files&relativeRef=%2Fmyresume%2Fdoc%3Fversion%3Dlatest
> ... where `myresume`, `doc`, `version`, and `latest` path elements might very reasonably use localized representations

w3cbot:
This was discussed during the [#did meeting on 25 September 2025](https://www.w3.org/2025/09/25-did-minutes.html#10000).

<details><summary><i>View the transcript</i></summary>

<h4>relativeRef should prefer UTF-8 for percent encoding <a href="https://github.com/w3c/did-resolution/issues/201" rel="noopener noreferrer">#201</a></h4>
<p><cite>ottomorac:</cite> next one is our dear friend Relative Ref</p>
[...]

Issue #200, opened 2025-09-22
DID parameters require ASCII-only


aphillips:
DID Parameters
https://www.w3.org/TR/did-resolution/#did-parameters

Various parameter's (`service`, `serviceType`, `versionId`, etc.) require that the value be an ASCII string. One parameter (`relativeRef`) also requires ASCII, but references RFC3986 section 2.1 (percent encoding). It is not clear why the values are restricted to ASCII. If the reason is so that they can traverse e.g. HTTP headers, then escaping should be supplied for non-ASCII values when such values are needed by the application. If there is some other reason, it is not apparent from the specification.

Some of the parameters, such as `versionTime` have a well-known, required serialization that is, in fact, ASCII only.

`service` seems to disadvantage systems that would otherwise permit non-ASCII service names.

`versionId` (and friends like `nextVersionId`) does not have a defined format. It's unclear why the ASCII restriction is imposed, except for (probably) the way that it simplifies the serialization story. Note that versioning is spelled out here:

* https://www.w3.org/TR/did-resolution/#versioning

ottomorac:
Thanks @aphillips I have created a PR using the language you have suggested.

Also for @msporny , I believe that once this PR is agreed to and merged, we should probably apply these same changes in the smaller set of DID parameters that are mentioned in DID Core here: https://www.w3.org/TR/did-1.0/#did-parameters  Perhaps we should create a mirror issue for this in DID Core?

PR #217, opened 2025-10-19
Prefer UTF-8 for DID Parameters

ottomorac:
This PR addresses #200 and #201 :
- Includes normative statement about UTF-8 for DID parameters
- Removes ASCII only normative statement from the individual DID parameters (except `versionTime` which should probably remain as ASCII only in my view)

***
<a href="https://pr-preview.s3.amazonaws.com/ottomorac/did-resolution/pull/217.html" title="Last updated on Oct 19, 2025, 3:45 AM UTC (349e2b6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-resolution/217/a0da1e3...ottomorac:349e2b6.html" title="Last updated on Oct 19, 2025, 3:45 AM UTC (349e2b6)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -325,6 +325,13 @@ did:example:123?service=files&amp;relativeRef=/resume.pdf
 	</pre>
 
 	<p>
+		For each DID parameter that is present, its associated value MUST be a 
+		<a data-cite="INFRA#scalar-value-string">scalar value</a>
+		 string serialized into ASCII according to <a data-cite="RFC3987#section-3.1">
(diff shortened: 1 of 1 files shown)


swcurran:
While not part of this PR (which I'm fine to merge), Line 324 -- immediately above the first change -- is incorrect, as the `relativeRef` query parameter is not percent-encoded as required.  Probably best to update that outside this PR, but it should be fixed.

Issue #218, opened 2025-10-20
Add rules for DID URL query normalization


msporny:
From the [SING review](https://github.com/w3c/security-request/issues/91#issuecomment-3420371725):
 
3. DID URL query normalization –  recommending enumeration of sharp edges for reviewers.   Retain and expand the current non-normative note. Enumerate edge cases such as:
* percent-encoding order and normalization,
* duplicate parameter handling,
* path traversal (../) issues, and
* canonicalization rules by DID Method

msporny:
A PR should be raised that clarifies the normalization rules around the edge cases above.

wip-abramson:
#301 merged, closing

Issue #209, opened 2025-10-07
Specify the algorithm for processing the hl DID parameter.


wip-abramson:
I noticed this TODO in the spec dereferencing algorithm - https://w3c.github.io/did-resolution/#dereferencing-algorithm-resource

We should either specify it or remove the hl DID parameter.

Anyone using the hl parameter able to write a PR for this?

msporny:
PR #262 has been raised to address this issue. This issue will be closed if that PR is merged.

Issue #181, opened 2025-08-17
Create syntactic sugar for `relativeRef` feature


msporny:
The current `relativeRef` feature is more awkward to use than necessary, resulting in DID URLs that look like this:

```did:example:123?service=agent&relativeRef=/images/123.png```

Which could result in this URL:

```https://domain.example/accounts/987/images/123.png```

We should introduce some syntactic sugar that makes the default use case for `relativeRef` generalized across DID Methods, so URLs like this could be used instead:

```did:example:123:/images/123.png```

The dereferencing algorithm would look for a service of type `UrlResolver` with a `baseUrl` property, which might look like this in the DID Document:

```
"service": [{
  "type": "UrlResolver",
  "baseUrl": "https://domain.example/accounts/987"
}]
```

The algorithm would then take the path parameters from the DID URL (`/images/123.png`), and then append them to the `baseUrl`, resulting in the final URL to dereference in order to retrieve the resource. DID Methods MAY provide additional or alternate rules for resolving URLs. Multiple base URLs MAY be provided for the purposes of redundancy.

**Comment by dlongley** [comment]
> Seems like a big price to pay for a single UrlResolver type.

Agreed, so if that would be the consequence of relying solely on a single type and existing protocols/patterns, then it seems it would be better to define a second type and simple protocol for URL translation for interacting with services of that type.

**Comment by peacekeeper** [comment]
Just to note from yesterday's DID WG call, `did:example:123:/images/123.png` isn't a valid DID URL per https://www.w3.org/TR/did-1.1/#did-url-syntax:

```
did-url = did path-abempty [ "?" query ] [ "#" fragment ]
```

and

```
did                = "did:" method-name ":" method-specific-id
method-name        = 1*method-char
method-char        = %x61-7A / DIGIT
method-specific-id = *( *idchar ":" ) 1*idchar
idchar             = ALPHA / DIGIT / "." / "-" / "_" / pct-encoded
pct-encoded        = "%" HEXDIG HEXDIG
```

@talltree may also have an opinion on this, he loves ABNF discussions :)

**Comment by msporny** [comment]
> For example, the service could express a base path on which it operates (one way or another), such that only if that base path is used in a DID URL will the resolver service will become operative (otherwise it will not).

Right, and so the following would become the example (which is conformant to DID URL syntax):

```
did:example:123/images/123.png
```

The dereferencing algorithm would look for a service of type UrlResolver with a baseUrl property, which might look like the following two examples. For a static resolver:

```
[...]

Issue #156, opened 2025-05-29
Define how DID Methods can specify paths and their dereferencing rules


wip-abramson:
Created based on Discussion in https://www.w3.org/2025/05/29-did-minutes.html

We would need to point to the DID Extensions and state how different DID methods can define the path and processing rules they are using.

There is also an ongoing discussion in #150 around whether paths should be processed consistently across DID methods.

msporny:
Also related: https://github.com/w3c/did-resolution/issues/181

w3cbot:
This was discussed during the [#did meeting on 21 August 2025](https://www.w3.org/2025/08/21-did-minutes.html#a824).

<details><summary><i>View the transcript</i></summary>

<h4>Define how DID Methods can specify paths and their dereferencing rules #156</h4>
<p><cite>&lt;ottomorac&gt;</cite> <a href="https://github.com/w3c/did-resolution/issues/156" rel="noopener noreferrer">w3c/<wbr>did-resolution#156</a></p>
<p><cite>&lt;ottomorac&gt;</cite> Issue created by Will based on a discussion from 29-May. We would need to point to the DID Extensions and state how different DID methods can define the path and processing rules they are using. This is also connected to issue #150 related to whether paths should be processed consistently across DID methods.</p>
<p><cite>Wip:</cite> Raised after discussion with Joe. Do we as a DID spec require a consistent experience across DID Methods, or is it a DID Method responsibility?</p>
<p><cite>&lt;manu&gt;</cite> <a href="https://github.com/w3c/did-resolution/issues/181" rel="noopener noreferrer">w3c/<wbr>did-resolution#181</a></p>
<p><cite>&lt;manu&gt;</cite> about a unified way to do path resolution</p>
<p><cite>manu:</cite> I think we should. Three similar issues about this.  Doesn't prevent DID Methods from having their own approaches, but we should try to get a consistent approach. Could be a long discussion, but without that we will get interop issues.</p>
<p><cite>JoeAndrieu:</cite> I think we need to do it. Markus had doubts, happy Manu is for it.  Should try to figure out, because there are lots of ways to do it, some in production. Differences in the DID Linked Resources, where is the data about the resource -- DIDDoc or Metadata or path?</p>
<p><cite>manu:</cite> Hope is that 181 proposal is not in conflict.  Special topic call to explore this as Joe suggested is a good idea.</p>
<p><cite>ottomorac:</cite> Chairs to discuss and set a deep-dive session time.</p>

<hr /></details>

PR #221, opened 2025-10-21
Update DID URL Dereferencing Section with Deterministic Path Handling

swcurran:
> **NOTE**: This PR is still in draft, as I need to clean up the internal section links, the definitions and references,
> anchors and links, and getting a consistent ReSpec style. I’m putting in the PR now to get feedback on the general ideas
> and approach vs. on the ReSpec nitty-gritty. As this is discussed, I’ll continue with the cleanup.
>
> A rendered version of the PR can be seen here: [https://swcurran.github.io/did-resolution/](https://swcurran.github.io/did-resolution/).

Addresses issues: #218, #209, #201, #200, #181, #156

This pull request significantly revises and expands the content of the **DID URL Dereferencing** section to clarify how
DID URL components — specifically the path, query parameters, and fragment — are to be interpreted during dereferencing.
The update draws an explicit analogy to HTTP URL handling, helping implementers relate familiar URL resolution behavior
to DID URLs, while also clearly defining where DID-specific semantics differ. Most notably, this change introduces
**deterministic and interoperable handling of the DID URL path**, which was previously undefined in the specification.
The additions establish a clear default behavior using the new `FileService` service type, while allowing for custom
behavior via method-specific or DID extension service types. The result is a specification that is easier to implement,
more extensible, and more aligned with the expectations of developers familiar with web infrastructure.

This update is (intended to be) **backwards compatible**:

* No existing required behavior has been changed or invalidated.
  * Techniques defined and used by Cosmos for Linked Resources and Cheqd for DID Linked Resources are untouched in their use of fragments and query parameters, respectively. The main focus here is on formalizing DID URL path handling.
* The `FileService` model supports DID Documents already using `serviceEndpoint` mappings that align with intuitive path-based routing.
* While the wording is updated, the handling of fragments and query parameters is (intended to be) unchanged.

**Summary of changes:**

* **Removed** the outdated **"DID Parameters"** section (now deprecated and unreferenced), and defining the Query Parameters defined in this spec to the “DID Resolution Query Parameters and Service Types” section.
* **Replaced** the previous **"DID URL Dereferencing Algorithm"** and **"Examples"** subsections with new normative and structured content.
*  **Added** new structured subsections to the **DID URL Dereferencing** section:
      * **DID URL Components: Path, Query Parameters, and Fragments** — defines interpretation rules for each component, with HTTP comparisons.
     * **DID URL Dereferencing Algorithm** — provides normative steps for resolving a DID URL to a resource via service mappings.
     * **DID Resolution Query Parameters and Service Types** — defines recognized query parameters (e.g., `service`) and registered service types (e.g., `FileService`).
        * This brings together into “mini-specifications” all of the query parameter and service types defined by the DID Resolution spec. The previous “DID Parameters” content is moved to this section.
    * **Examples** — includes multiple concrete cases showing default and custom behaviors, including a fictitious `FictitiousVerifiedFileService`.

***
<a href="https://pr-preview.s3.amazonaws.com/swcurran/did-resolution/pull/221.html" title="Last updated on Oct 28, 2025, 7:01 PM UTC (f14b1ac)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-resolution/221/a0da1e3...swcurran:f14b1ac.html" title="Last updated on Oct 28, 2025, 7:01 PM UTC (f14b1ac)">Diff</a>


index.html
```diff
@@ -1246,334 +1093,764 @@ <h3>DID URL Content Metadata</h3>
...
-						<li><b>dereferencingMetadata</b>: <var>error object</var> with type set to <a href="#INVALID_DID_URL">INVALID_DID_URL</a></li>
-						<li><b>contentStream</b>: <code>null</code></li>
-						<li><b>contentMetadata</b>: <code>«[ ]»</code></li>
+		<p>
+			A <a>DID URL</a> consists of a <a>DID</a>, optionally followed by
+			additional URL components — a path, query parameters, and a fragment
+			— as defined in
+			<a
+			href="https://www.rfc-editor.org/rfc/rfc3986#section-3">RFC3986</a>.
+			This section defines how these components <strong>MUST</strong> be
```
```suggestion
			This section defines how these components MUST be
```

I don't think we want to bold normative statements to maintain consistency with the rest of the spec text.

wip-abramson:
> Addresses issues: https://github.com/w3c/did-resolution/issues/218, https://github.com/w3c/did-resolution/issues/209, https://github.com/w3c/did-resolution/issues/201, https://github.com/w3c/did-resolution/issues/200, https://github.com/w3c/did-resolution/issues/181, https://github.com/w3c/did-resolution/issues/156

Thanks @swcurran!

I noticed that the DID Resolution Options section still references #did-parameters for its definitions of versionId and versionTime. Perhaps we just point them to the new section these definitions moved to?

index.html
swcurran: Thanks @wip-abramson -- will remove.  As noted, I have a pure cleanup pass still to go. I should have checked on this one....

index.html
```diff
@@ -1246,334 +1093,763 @@ <h3>DID URL Content Metadata</h3>
...
+						<li><code>contentMetadata</code>: metadata about the
+						resolved DID document</li>
+						<li><code>dereferencingMetadata</code>: any metadata
+						about the dereferencing process</li>
+					</ul>
+				</li>
+				<li>
+					If the resolution result is a resource resolved via DID URL path processing, return:
+					<ul>
+						<li><code>contentStream</code>: the resolved
```
dmitrizagidulin: Are you sure we want to inline the content stream into the resulting json object? What if it's like a large video or image?

dmitrizagidulin (approved):
Amazing work 👍 

(One minor question about the contentStream of the resolution result)

swcurran:
Added a few more comments, cleaning up the links in the document, making sure all links that are expected are there, and trying to be consistent with the formatting of the HTML. I think I will need to do one more full pass before I remove the "Draft" status, so I don't waste people's time on this.

Looking forward to discussing this at the DID Working Group meeting.

ottomorac:
DID WG Discussion on 23-Oct:

https://www.w3.org/2025/10/23-did-minutes.html

peacekeeper (changes requested):
I think the general idea is interesting, and some years ago I also used to think about the DID Path in the same way. But today I have several concerns with this PR..

1. I don't really like how this completely replaces the current "DID URL Dereferencing Algorithm" section. It would be better if this updates or extends the current section to add the new functionality. For example, some of the effects of replacing the current section would be:

    - In the current algorithm, it's possible to have just a "service" parameter without path and without "relativeRef". The PR would break this.
    - The current algorithm defines how to process the "verificationRelationship" option. The PR completely removes this, why?
    - Some error handling details are also being removed here.

2. Some statements in the PR are clearly wrong, such as "A DID URL MUST include a path component".

3. As discussed, I think this breaks how some existing DID methods use the path, e.g. under the current rules, `did:cheqd:testnet:55dbc8bf-fba3-4117-855c-1e0dc1d3bb47/resources/398cee0a-efac-4643-9f4c-74c48c72a14b` is dereferenced to an image resource, without depending on service endpoints, and without requiring the use of HTTPS endpoints. The PR would break this.

4. If we decide that we want this functionality in DID Resolution, then I think it would be important to NOT allow DID methods to override it, so that resolvers and clients can be certain that this feature always works in the same way, just like "service" and "relativeRef" work today in a completely DID method independent way. So for me, it would be important to know how the `did:cheqd` community feels about this, and if they could be convinced to abandon their current use of the DID Path.

5. I think this PR also breaks a feature in the existing HTTPS Binding section, which can return a 303 redirect to a selected service endpoint URL.

6. Note that in the current algorithm, the result of dereferencing a DID URL with a "service" / "serviceType" and optional "relativeRef" parameter is a list of selected service endpoint URLs. The result is NOT a resource at those URLs. I believe it has been our common understanding for a while that retrieving something from a service endoint URL is a separate dereferencing step that happens after DID URL dereferencing.

7. I don't think the way how path and query parameters are used together is conformant with basic URI architecture. In an example like ` did:example:123456789abcdefghi/path/to/file.txt?versionTime=2025-10-01T12:00:00Z`, I would expect that "versionTime" means the version of the `file.txt`, not the version of the DID document. Furthermore, I think there's some kind of layering violation, if this PR says that query parameters "understood by the DID resolver" are used for the DID document, and that "unused query parameters" are appended to a service endpoint URL. The results could differ by resolver implementation and therefore be unpredictable.

    - Just as a side note, this is why matrix parameters would have been nice. Then we could have cleanly separated "parameters for the DID document" and "parameters for the service endpoint URLs". E.g. the example above would have looked like ` did:example:123456789abcdefghi;versionTime=2025-10-01T12:00:00Z/path/to/file.txt?otherparam=value`. Note that the introduction of `relativeRef` was actually a direct result of removing matrix parameters, see @dlongley 's comment https://github.com/w3c/did/issues/159#issuecomment-598286814 and issue https://github.com/w3c/did/issues/349

Finally, I'm sorry, but I'm still not sure if I agree with the motivation here. I think the fact that path handling is left completely to the DID method can also be considered a feature, not a problem. The current design means that DID URLs with paths provide a namespace that fully inherits all the benefits of DIDs, such as decentralization, persistence, cryptographic verifiability. Using a DID Path and using "relativeRef" serve two different purposes. This PR would abandon this design by binding the DID Path to service endpoints, which in practice will be HTTPS URLs and therefore have a lot of potential downsides that DIDs were designed to solve in the first place.

swcurran:
I'll add in-line comments:

> I think the general idea is interesting, and some years ago I also used to think about the DID Path in the same way. But today I have several concerns with this PR..
> 
> 1. I don't really like how this completely replaces the current "DID URL Dereferencing Algorithm" section. It would be better if this updates or extends the current section to add the new functionality. For example, some of the effects of replacing the current section would be:
>    

The DID URL Dereferencing algorithm [remains](https://swcurran.github.io/did-resolution/#dereferencing-algorithm). It has been adjusted to reflect the guidance provided in the "Components Handling" section, and to encourage that handling.

>    * In the current algorithm, it's possible to have just a "service" parameter without path and without "relativeRef". The PR would break this.

That could still be valid by including that in the definition of "service". I would think that the behaviour is better accomplished by using a fragment that is the `id` of the service. That said, by using query parameters `service` and `serviceType` without a path, I'm not sure what the result is -- a guess the DIDDoc with an array of "node references" to services -- as if there were multiple fragments.  I'm also not sure that would be useful. 

>    * The current algorithm defines how to process the "verificationRelationship" option. The PR completely removes this, why?

I'll look at that -- it was not intended to change.

>    * Some error handling details are also being removed here.

Happy to include them back in. I'll see what I can determine. I've never been very clear on when it is OK to say "report an error" vs. "report a specific error" in the DID family of specs. AFAIK, only "NOT FOUND" and "ERROR" are defined.

> 2. Some statements in the PR are clearly wrong, such as "A DID URL MUST include a path component".

I think you were looking at an earlier version of what I submitted -- not the latest commit. When I removed the `<STRONG>` tags I accidentally changed some MAYs to MUSTs (doh!!!), but in the last cleanup, those were all changed. If any are still there, that is clearly an error. Sorry about that.

> 3. As discussed, I think this breaks how some existing DID methods use the path, e.g. under the current rules, `did:cheqd:testnet:55dbc8bf-fba3-4117-855c-1e0dc1d3bb47/resources/398cee0a-efac-4643-9f4c-74c48c72a14b` is dereferenced to an image resource, without depending on service endpoints, and without requiring the use of HTTPS endpoints. The PR would break this.

As discussed on the call, I think we settled that I would add a step to say "DID Methods specific handling" can be done in the process, but is discouraged.

Without that, the [DID Linked Resources](https://w3c-ccg.github.io/did-linked-resources/) spec will break. That said, the goal of DID Linked Resources -- especially for Cheqd -- would have been far easier and logical with this PR in place. They would have simply defined a Cheqd `service` type that used the path `/resources` to be selected, extracted the `collectionId` from the DID and the `resourceId` from the path. So yes, some things may be different, but I think it makes future work much easier.

> 4. If we decide that we want this functionality in DID Resolution, then I think it would be important to NOT allow DID methods to override it, so that resolvers and clients can be certain that this feature always works in the same way, just like "service" and "relativeRef" work today in a completely DID method independent way. So for me, it would be important to know how the `did:cheqd` community feels about this, and if they could be convinced to abandon their current use of the DID Path.

Fully agree with that.  Let's have the conversation.

> 5. I think this PR also breaks a feature in the existing HTTPS Binding section, which can return a 303 redirect to a selected service endpoint URL.

I'll take a look at that, but could probably use some help in understanding that.

> 6. Note that in the current algorithm, the result of dereferencing a DID URL with a "service" / "serviceType" and optional "relativeRef" parameter is a list of selected service endpoint URLs. The result is NOT a resource at those URLs. I believe it has been our common understanding for a while that retrieving something from a service endoint URL is a separate dereferencing step that happens after DID URL dereferencing.

This comment is HUGE.  My understanding is that a DID Path (even one produced by `relativeRef`) is expected to return the resource, not an array of URLs. If DID with path is not expected to return the referenced resource, but rather the URL for the expected resource, we have a big misunderstanding.  I think it should return the resource -- hence this PR. Definitely something that needs to be clarified.

> 7. I don't think the way how path and query parameters are used together is conformant with basic URI architecture. In an example like ` did:example:123456789abcdefghi/path/to/file.txt?versionTime=2025-10-01T12:00:00Z`, I would expect that "versionTime" means the version of the `file.txt`, not the version of the DID document. Furthermore, I think there's some kind of layering violation, if this PR says that query parameters "understood by the DID resolver" are used for the DID document, and that "unused query parameters" are appended to a service endpoint URL. The results could differ by resolver implementation and therefore be unpredictable.

This is why I wanted the "Query Parameters and Service Types" section to be a series of "mini-specs" to avoid exactly this issue.  I understand your comment, but I disagree with it as applied to DIDs. In what is above, the `versionTime` parameter can only be used without a path -- with, I guess, the weird exception of the path being to a DID. But an important capability of DIDs is that we can achieve "long lasting" identifiers by being able to reference the DID at a particular version. For example, if the `/path/to/file.txt` has a related data integrity proof (perhaps it is a VC) that is based on the key in the particular version of a DID active at the time the file was created it is necessary to have a way to find both the key and the resource. I think that is a crucial feature of DIDs -- long lasting verifiability, while also allowing for best practices key handling.

An alternate that we discussed to clarify this is that there could be a query parameter on the DID that specified (in percent encoded form) that query parameters to be included in the assembled URL. That way, any query parameters not understood can be thrown away and there are ONLY DID Resolver query parameters.

As an aside on this point -- I would also like to see added to the DID Spec the capability to clearly deactivate keys used in prior versions of a DIDDoc with an indicator such as "we've rotated to a new key and is no longer being used" and "this key was compromised -- don't trust anything signed with it". We have deactivating a DID, but I think we should also focus on deactivating keys while retaining the DID.

>    
>    * Just as a side note, this is why matrix parameters would have been nice. Then we could have cleanly separated "parameters for the DID document" and "parameters for the service endpoint URLs". E.g. the example above would have looked like ` did:example:123456789abcdefghi;versionTime=2025-10-01T12:00:00Z/path/to/file.txt?otherparam=value`. Note that the introduction of `relativeRef` was actually a direct result of removing matrix parameters, see @dlongley 's comment [w3c/did#159 (comment)](https://github.com/w3c/did/issues/159#issuecomment-598286814) and issue [w3c/did#349](https://github.com/w3c/did/issues/349)

I don't think this is necessary, and would be more complex. It's also not AFAIK compliant with RFC3986. But that's just me.  :-). Happy to discuss.

> 
> Finally, I'm sorry, but I'm still not sure if I agree with the motivation here. I think the fact that path handling is left completely to the DID method can also be considered a feature, not a problem. The current design means that DID URLs with paths provide a namespace that fully inherits all the benefits of DIDs, such as decentralization, persistence, cryptographic verifiability. Using a DID Path and using "relativeRef" serve two different purposes. This PR would abandon this design by binding the DID Path to service endpoints, which in practice will be HTTPS URLs and therefore have a lot of potential downsides that DIDs were designed to solve in the first place.

In writing the PR I was as focused as much on Cheqd's "resources on a blockchain", on "resources as files secured on a blockchain", and on "IPFS" use cases as I was with "resources with HTTPS URLs". I tried to make sure all were enabled.  I think the use of a service `type` to define how to handle a path is extensible such that the goal can be accomplished, with the ability of service types defined at the DID resolution spec level (all DID Methods), the DID Method level (one DID method -- not ideal) and the DID Extensions Registry (multiple DID Methods) is a good balance. Some things like using IPFS can be accomplished using the DID Resolution spec-defined `FileServce` (assuming the DID Resolver knows how to resolve such a URL), simply by using that protocol in the `serviceEndpoint`.

Let's discuss further.

swcurran:
DID WG Discussion on 23-Oct, I think the following changes were requested:

- Update to reference a "DID dereferencer" instead of "DID resolver" where appropriate.
- Add the (discouraged, but possible) "DID Method-specific" path handling step to the processing flow.
- Add "collect types" step in the process.

I'll also reread the cheqd DID Linked Resources Spec and think about how to get it to align with the "new" DID Resolution approach for paths.

index.html
```diff
@@ -1246,334 +1098,787 @@ <h3>DID URL Content Metadata</h3>
...
-					<a href="https://www.w3.org/TR/did-core/#did-url-syntax">DID URL Syntax</a>.
-					If not, the <a>DID URL dereferencer</a> MUST return the following result:
-					<ol class="algorithm">
-						<li><b>dereferencingMetadata</b>: <var>error object</var> with type set to <a href="#INVALID_DID_URL">INVALID_DID_URL</a></li>
-						<li><b>contentStream</b>: <code>null</code></li>
-						<li><b>contentMetadata</b>: <code>«[ ]»</code></li>
+		<p>
+			A <a>DID URL</a> consists of a <a>DID</a>, optionally followed by
+			additional URL components — a path, query parameters, and a fragment
+			— as defined in
```
```suggestion
			additional URL components — any combination of a path, zero or more query
			parameters, and/or a fragment identifier — as defined in
```

index.html
```diff
@@ -1246,334 +1098,787 @@ <h3>DID URL Content Metadata</h3>
...
+			retrieving related resources).
+			</p>
+		</section>
+
+		<section id="did-parameters">
+			<h4>Query Parameters</h4>
+
+			<p>
+			A <a>DID URL</a> MAY include a single query
+			component, as defined in
```
```suggestion
			A <a>DID URL</a> MAY include one or zero
			query component(s), which, if present, MAY include
			zero or more query parameters, as defined in
```

index.html
```diff
@@ -1246,334 +1098,787 @@ <h3>DID URL Content Metadata</h3>
...
+			parameters appear after the DID and any path component, and their
+			structure and encoding MUST follow standard URL syntax conventions.
+			Their interpretation depends on the context in which the DID URL is
+			used and can influence the dereferencing result.
+			</p>
+
+			<p>
+			Unlike HTTP query parameters, which are interpreted by customized web server
+			implementations, <a href="#did-resolution-parameters">DID parameters</a> in DID
+			URLs MUST have defined semantics. A <a>DID Resolver</a> MUST process
```
```suggestion
			URLs MUST have predefined semantics. A <a>DID Resolver</a> MUST process
```

index.html
```diff
@@ -1246,334 +1098,787 @@ <h3>DID URL Content Metadata</h3>
...
+			only recognized query parameters — specifically those defined in
+			this specification (see the <a href="#did-resolution-parameters">DID Parameter
+			definitions</a>), the applicable <a>DID method</a> specification, or
```
```suggestion
			definitions</a>), the applicable <a>DID method</a> specification, and/or
```

index.html
```diff
@@ -1246,334 +1098,787 @@ <h3>DID URL Content Metadata</h3>
...
+			DID Document Properties Extensions [[?DID-EXTENSIONS-PROPERTIES]].
+			All other query parameters MAY be passed through unchanged for
+			downstream processing, such as <a>dereferencing</a> a
+			<a>DID URL</a> path.
+			</p>
+
+			<p>
+			Query parameters may affect dereferencing behavior, such as
+			selecting a versioned resource or influencing how a path is
```
```suggestion
			Query parameters may affect dereferencing behavior, such as by
			specifying a versioned resource or by influencing how a path is
```

index.html
```diff
@@ -1246,334 +1098,787 @@ <h3>DID URL Content Metadata</h3>
...
+			to be processed. The impact of specific query parameters defined in this
+			specification in combination with a <a>DID URL</a> path is described
+			in the <a href="#did-url-path">next</a> section of this.
```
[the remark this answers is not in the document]
TallTed: or
```suggestion
			in the <a href="#did-url-path">next section</a>.
```

index.html
```diff
@@ -1246,334 +1098,787 @@ <h3>DID URL Content Metadata</h3>
...
-						<ol class="algorithm">
-							<li><b>dereferencingMetadata</b>: <var>error object</var> with type set to <a href="#NOT_FOUND">NOT_FOUND</a></li>
-							<li><b>contentStream</b>: <code>null</code></li>
-							<li><b>contentMetadata</b>: <code>«[ ]»</code></li>
-						</ol>
+				<li>Extract the path component from the <a>DID URL</a>.</li>
+				<li>
+					Match the extracted path against the <code>prefix</code>
+					attribute of each relevant <code>service</code> entry in the
+					DID Document that includes a <code>prefix</code> attribute.
```
```suggestion
					DID Document that includes such an attribute.
```

index.html
```diff
@@ -1246,334 +1098,787 @@ <h3>DID URL Content Metadata</h3>
...
+					If the <code>service</code> and/or <code>serviceType</code>
+					<a href="#did-resolution-parameters">DID Parameters</a> are
+					present, the <a>DID Resolver</a> MUST consider only services
+					with the specified <code>id</code> or <code>type</code>
+					(respectively) values. Otherwise, all
```
```suggestion
					with the respectively specified <code>id</code> and/or
					<code>type</code> values. Otherwise, all
```

index.html
```diff
@@ -1246,334 +1098,787 @@ <h3>DID URL Content Metadata</h3>
...
-						<li>Otherwise, return the following result:
-							<ol class="algorithm">
-								<li><b>dereferencingMetadata</b>: <var>error object</var> with type set to <a href="#REPRESENTATION_NOT_SUPPORTED">REPRESENTATION_NOT_SUPPORTED</a></li>
-								<li><b>content</b>: <code>null</code></li>
-								<li><b>contentMetadata</b>: <code>«[ ]»</code></li>
-							</ol>
+				<li>
+					Use the <code>type</code> of the selected
+					<code>service</code> to determine how the
+					<a>DID URL</a> path is used in assembling the URL of the
```
```suggestion
					<a>DID URL</a> path is to be used in assembling the URL of the
```

index.html
```diff
@@ -1246,334 +1098,787 @@ <h3>DID URL Content Metadata</h3>
...
+						<p>
+						Resolving an assembled URL — particularly one that is a
+						DID — might result in a <dfn>resolution cycle</dfn>,
+						which is a set of steps that result in an infinite loop.
+						For example, a <a>service endpoint</a> might indirectly
+						point back through a sequence of resolutions to a
+						previously dereferenced identifier. A <a>DID
+						resolver</a> recursively resolving a <a>service
+						endpoint</a> is advised to detect and handle such a
+						cycle to prevent an infinite loop or resolution failure.
```
```suggestion
						previously dereferenced identifier. A <a>DID
						resolver</a> is expected to detect and handle
						recursive cycles when resolving a <a>service
						endpoint</a>, to prevent an infinite loop or
						other resolution failure.
```

index.html
```diff
@@ -1246,334 +1098,787 @@ <h3>DID URL Content Metadata</h3>
...
+
+			<li>
+				<strong>Handle Fragments:</strong>
+				<ul>
+					<li>As noted in <a href="#did-url-fragments">DID URL
+					fragments handling</a> section, if the <a>DID URL</a>
+					contains a fragment, a <a>DID Resolver</a> SHOULD NOT process
+					or interpret it. Fragment handling is the client's
+					responsibility given the resolved document and media
+					type.</li>
```
```suggestion
					<li>As noted in the <a href="#did-url-fragments">DID URL
					fragments handling</a> section, if a <a>DID URL</a>
					contains a fragment, a <a>DID Resolver</a> SHOULD NOT process
					or interpret it. Fragment handling is the client's
					responsibility, based on the resolved document and its media
					type.</li>
```

index.html
```diff
@@ -1246,334 +1098,787 @@ <h3>DID URL Content Metadata</h3>
...
+					<code>qp</code> working set.</li>
+				</ul>
+			</li>
+
+			<li>
+				<strong>No Path Component:</strong>
+				<ul>
+					<li>If no <code>path</code> is present, resolution is
+					complete and the path-handling step is skipped.</li>
+				</ul>
```
TallTed: One-bullet lists generally shouldn't be lists.
```suggestion
				If no <code>path</code> is present, resolution is
				complete and the path-handling step is skipped.
```

index.html
swcurran: Shouldn't this be "...if present, MAY include one or more query parameters..."

peacekeeper:
> > 1. I don't really like how this completely replaces the current "DID URL Dereferencing Algorithm" section. It would be better if this updates or extends the current section to add the new functionality. For example, some of the effects of replacing the current section would be:
> 
> The DID URL Dereferencing algorithm [remains](https://swcurran.github.io/did-resolution/#dereferencing-algorithm). It has been adjusted to reflect the guidance provided in the "Components Handling" section, and to encourage that handling.

I think it would be fine if this PR actually "adjusts" the existing DID URL Dereferencing algorithm, by adding the proposed functionality. But instead it completely replaces the entire content about DID URL Dereferencing. This causes a lot of issues with other functionality and other parts of the specification.

I have already listed some of the issues in my previous content, but this PR includes major changes that I think will really affect many other parts of the document. For example, some statements in the "DID Resolution Architectures" section would also be broken by this PR.

> > Removed the outdated "DID Parameters" section

I don't understand this either. A PR that proposes changes to how dereferencing of the DID Path is handled shouldn't have to suddenly declare the entire existing content about DID Parameters "outdated" and "deprecated".

wip-abramson:
> > 1\. Note that in the current algorithm, the result of dereferencing a DID URL with a "service" / "serviceType" and optional "relativeRef" parameter is a list of selected service endpoint URLs. The result is NOT a resource at those URLs. I believe it has been our common understanding for a while that retrieving something from a service endoint URL is a separate dereferencing step that happens after DID URL dereferencing.
> 
> This comment is HUGE. My understanding is that a DID Path (even one produced by `relativeRef`) is expected to return the resource, not an array of URLs. If DID with path is not expected to return the referenced resource, but rather the URL for the expected resource, we have a big misunderstanding. I think it should return the resource -- hence this PR. Definitely something that needs to be clarified.
> 
> >

The current dereferencing algorithm defined in the spec appears to be aligned with what @peacekeeper states, specifically steps 8.2, 8.3 and 8.4 of the derefencing a resource algorithm - https://w3c.github.io/did-resolution/#dereferencing-algorithm-resource.

> My intention was to "adjust" the algorithm.

It does seem that the algorithm has been rewritten from scratch - I appreciate that the intention was to keep the same functionality as the previous version. 

As there seems to be a lot for the WG to discuss here, we will have a special topic call on Wednesday the 5th of November dedicated to this PR. In the meantime, lets continue to conversation here. Especially interested to hear from @tweeddalex and @jandrieu

tweeddalex:
Hi @swcurran coming to this fresh and have a lot to catch up on! I'll caveat the below comments by saying that I probably need to read through the proposed updates again:

- Am I correct to assume that the `FileService` type would be much more valuable for **web-based DID methods** than other types of DID Methods?
- With web-based DID methods, the `FileService` type seemingly helps resolvers negotiate request formats into coherent HTTP URLs, pointing to specific endpoints or resources. This makes sense for `did:web` because the HTTP URL can be extracted from the DID and appended with the contents of the prefix/service endpoint. 
- However, for non-web DID Methods, I don't see universal value. E.g. if we added a `/resources/` prefix and `FileService` into `did:cheqd` DID Documents, a resolver or DID URL Dereferencer would still need to be aware of how to recompile DID Documents and Resources from the cheqd ledger. 
- Therefore, on first read - I see this being useful for a sub-set of DID Methods (that can be wrangled into HTTP URLs), rather than a blanket `service` type that should be universally adopted by ALL resolvers and DID URL dereferencers. 
- Perhaps this could be its own standalone spec for applying to `did:web`, `did:webvh`, `did:webs`, etc.

ankurdotb:
## Context

This suggested edit does seem to break the DID Linked Resources specification.

> That said, the goal of DID Linked Resources -- especially for Cheqd -- would have been far easier and logical with this PR in place. They would have simply defined a Cheqd service type that used the path /resources to be selected, extracted the collectionId from the DID and the resourceId from the path. So yes, some things may be different, but I think it makes future work much easier.

To provide a bit of motivation behind why DID-Linked Resources was created and to answer some of the points raised above:

1. The reason we didn't go with the `service` section is that many DID Core authors, including Drummond, clarified that the intention of the `service` section was to point to APIs, endpoints etc and not to specific files. The other consideration is we wanted to add richer metadata on versioning, file media type, versioning etc which would be hard to add within the `service` section without deviating from the definition in DID Core (see below on that).
2. Adding a new top-level section in the DID Document was considered, and then not selected because it could not be standardised as core property without modifying DID Core (which was considered tough to achieve). While the DID spec does allow new properties to be added, practically speaking many software libraries will have strong typing/validation to only parse the core properties. So while this change could be made at the _specification_ level, it might break practical implementations.
3. Hence, the approach adopted was to consider linked resource resolution as a subset of DID Resolution rather than DID Core, since in practical terms a lot of software libraries treat typing/validation more permissively in the DID Document metadata than in core, and would also the specification to be treated for standardisation with the DID Resolution spec.

(Sidebar: @tweeddalex The current spec draft for DLRs [does call out some of architectural considerations](https://w3c-ccg.github.io/did-linked-resources/#serviceendpoint), but we should perhaps expand these considerations to explain some of the above, which I do recall being answered/shown in decks about the spec.)

## Suggested path forward

I don't per se disagree with the idea of a `FileService` service type. I think that approach fits within the semantic definition of "service types define type of API endpoint". IMO, the architectural tradeoff this model makes is not having as rich metadata to present as query parameters but that's an okay tradeoff for a DID method to make. Especially in web-based DID methods, such file path handling can be done using HTTP headers like media type, `ETag` to understand cache expiry etc.

What I do think would be a less invasive way to do this would be to add a `FileService` type in DID Spec Registries, and defining how it behaves there, rather than changing the DID Resolution spec so extensively. The concern identified on breaking DLR spec is just one topic (and there are 6-7 other DID methods besides `did:cheqd` that implement DLR spec). There might be other, unforseen breakages that we have not identified in this thread. So I would suggest:

1. Define `FileService` service type in DID Spec Registries, along with some basic, DID method agnostic resolution guidance
2. Provide detailed guidance on how `FileService` service type resolution is done for a specific DID method, e.g., `did:webvh`
3. Keep DID Resolution spec unchanged, except for any **minor**, non-breaking edits.

swcurran:
Responding to @tweeddalex comment [here](https://github.com/w3c/did-resolution/pull/221#issuecomment-3460429442) and his fundamental question here *"Am I correct to assume that the `FileService` type would be much more valuable for **web-based DID methods** than other types of DID Methods?"*

The short answer is no :-) -- definitely not my intention. If it is, that needs to be fixed.  In writing this response, I think I see something in what I've done that could be adjusted to make it more inclusive -- covered in the **PROPOSAL** below. 

Here is my motivation on this PR:

The core issue is to define a useful, standard and flexible way to handle paths in a DID URL. Currently, step 9 of the  [Dereferencing the resource](https://w3c.github.io/did-resolution/#dereferencing-algorithm-resource) algorithm says essentially "path handling is not defined by the DID Resolution spec". It is left to the DID method, extensions or the client (??) to decide. I think that makes it really hard for everyone to use DID URL paths and I think that is unfortunate. DID URLs with paths are extremely useful, as almost all of us have found that want to associate resources with DIDs.

This PR attempts to fill that gap with the following approach.

1. When you have a path, find an appropriate service for intended for that path, with a [definition of how to find the service](the [path matching algorithm](https://swcurran.github.io/did-resolution/#did-url-path)).
2. Given the service found, use the specification for the type of the service to understand the rules to assemble a URL from the path. The definitions for the service types can be found in the DID Resolution spec, the spec for the DID Method, and the DID Extensions.
3. Resolve the assembled URL.

The [proposed specification for the `FileService` service type is in the DID Resolution spec](https://swcurran.github.io/did-resolution/#fileservice-service-type), and it is used to combine a base URL with the path.  It is useful for any resource that can has a URL -- regardless of the DID Method.

This should give two levels of flexibility:

1. If the meaning of the path can be expressed as a URL (HTTPS, IPFS, DID, MAGNET, etc.) then FIleService suffices.
2. If not, specify another Service Type that can be used at the DID Resolution spec level, the DID Method or the DID Extensions level.

**PROPOSAL** As I've responded to this I realize that my proposed 3 steps (above) should really be 2, with the 3rd step ("Resolve the URL") removed and left to the discretion of the specification of the service type (Step 2).  I've been assuming that Step 2 produces a URL based on the service type, Step 3 resolves that URL. Instead it should be Step 2 -- "process the path according to the service type definition to produce a resource".  For the "FileService" service type, that means assembling a URL and resolving it, but other service types can do other things without needing to produce a URL.

This would make Step 2:

2. Given the service found, use the specification for the type of the service to process the path and return a resource.

peacekeeper:
> My understanding is that a DID Path (even one produced by relativeRef) is expected to return the resource, not an array of URLs.

> Given the service found, use the specification for the type of the service to process the path and return a resource.

As @wip-abramson pointed out, this is not what the current algorithm does. It can return a DID document with the selected services, or list of selected service endpoint URIs, depending on the "accept" option.

A DID resolver returning an array of selected service endpoint URIs is comparable to a web server returning a 303 redirect. The client can then choose to dereference the other URI(s) or not. What you are suggesting would be similar to saying that a web server should contact another web server at a redirected URI, fetch the resource from there, and return it to the client. You would be combining two dereferencing processes into one, without involving the client in dereferencing the second URI.

If I read your PR correctly, the "service" parameter wouldn't even be processed unless there is a DID Path. Dereferencing something like `did:example:123?service=agent` would be undefined. Also, with your PR, only one "service" could be selected, whereas the current algorithm supports selecting multiple services (which can be a common use case, especially when `serviceType` is used).

> Is there a philosophical objection to defining in the DID Resolution spec a consistent way for how a path is to be processed?

I don't think you're actually achieving this, since with your proposal, the result of dereferencing the DID Path would depend on a mix of DID method specific logic AND on the contents of the DID document. Sometimes dereferencing the path would depend on the services in the DID document, and sometimes it wouldn't ("If the DID Method specifies its own path handling algorithm, it MUST be applied, and the rest of this algorithm skipped.").

Compared to this, in the current algorithm you have:
- "service" and "relativeRef" always work in a consistent, DID-method independent way. No DID method overrides this or defines any custom method-specific logic for this.
- DID Path is left entirely to the DID method, without any interference from the DID Resolution spec.

Put differently, today the DID Path is essentially an addressing mechanism WITHIN the DID method's VDR. In the case of did:cheqd, the DID Path is used to identify resources within the Cheqd VDR. In the case of web-based DID methods, the DID Path should be used to identify resources within the web server.

What you are suggesting is that the DID Path would be used to identify resources anywhere outside the DID method's VDR, determined by services. E.g. with your proposal, something like `did:web:example.com/path/resume.pdf` could return a PDF resource from anywhere outside of example.com, which seems very strange to me. With the current approach, the "did:web" method could simply specify that `did:web:example.com/path/resume.pdf` returns the resource at `https://example.com/path/resume.pdf`, which I think makes much more sense.

swcurran:
@peacekeeper , I have a some comments to make in response to your [last comment](https://github.com/w3c/did-resolution/pull/221#issuecomment-3470869908) that I'll put in after, but the most important one is to ask how the following is supposed to work -- via the DID Resolution spec: 

> With the current approach, the "did:web" method could simply specify that did:web:example.com/path/resume.pdf returns the resource at https://example.com/path/resume.pdf

Is there something in the current DID Resolution spec that says that is supposed to be how it is happen?

peacekeeper:
> Is there something in the current DID Resolution spec that says that is supposed to be how it is happen?

In this section: https://www.w3.org/TR/did-resolution/#dereferencing-algorithm-resource

Step 9.1 says:

> The applicable [DID method](https://www.w3.org/TR/did-resolution/#dfn-did-method-s) MAY specify how to dereference the [DID path](https://www.w3.org/TR/did-resolution/#dfn-did-paths) and/or [DID query](https://www.w3.org/TR/did-resolution/#dfn-did-queries) of the input [DID URL](https://www.w3.org/TR/did-resolution/#dfn-did-urls).

This means that the `did:web` DID method (and potentially other web-based DID methods) could define how to dereference something like `did:web:example.com/path/resume.pdf`

peacekeeper:
@swcurran I'm not fundamentally opposed to this. As I said, several years ago I also used to think about the DID Path exactly like you do now. But the removal of matrix parameters and simultaneous introduction of relativeRef changed a lot about the mechanics of DID URLs including the DID Path.

If we can answer some of the open questions in this thread, especially about backwards compatibility, DID URLs that don't require HTTP endpoints, and about the interpretation of DID parameters, then maybe there's a way to make it work. I just don't see it yet in this current PR, which in my opinion breaks a lot of things.

Issue #233, opened 2025-10-24
Avoid ambiguous use of 'match' in specification


wip-abramson:
> "match" is a yellow flag in specifications. Does this mean that it must be the same string, or does it use some other matching algorithm?

See https://w3c.github.io/did-resolution/#resolving

We mean string equivalence

wip-abramson:
Consensus to change this to string equivalence

Issue #231, opened 2025-10-24
didDocument definition should refer to DID Data Model


wip-abramson:
> It would be ideal if the definition of didDocument referred to https://www.w3.org/TR/did-1.0/#data-model, which actually defines DID documents.

From https://github.com/w3ctag/design-reviews/issues/1157#issuecomment-3439062470

(I think we should probably use 1.1).

PR #249, opened 2025-12-01
Update didDocument definition to remove ambiguous use of match and reference the DID data model section

wip-abramson:
This addresses #231 and #233.

The only part that is missing is the references are still going to DID 1.0 instead of 1.1. I think this is something to do with ReSpec and the fact that DID 1.1 is not a full standard yet. As soon as it is, I believe these references would update?

***
<a href="https://pr-preview.s3.amazonaws.com/wip-abramson/did-resolution/pull/249.html" title="Last updated on Dec 1, 2025, 1:40 AM UTC (4953f62)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-resolution/249/a0da1e3...wip-abramson:4953f62.html" title="Last updated on Dec 1, 2025, 1:40 AM UTC (4953f62)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -540,10 +540,10 @@ resolve(did, resolutionOptions) →
 		<dd>
 			If the resolution is successful, this MUST be a <a>DID document</a> that
 			is capable of being represented in one of the conformant
-			<a data-cite="did-core#representations">representations</a> of the [[[DID]]] specification.
-			The value of <code><a data-cite="did-core#dfn-id">id</a></code> in the resolved <a>DID document</a> MUST
-			match the <a>DID</a> that was resolved. If the resolution is unsuccessful, this
-			value MUST be empty.
+			<a data-cite="did#representations">representations</a> of the <a data-cite="did#data-model"></a> 
+			specification. The value of <code><a data-cite="did#dfn-id">id</a></code> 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -536,10 +540,10 @@ <h1>DID Resolution</h1>
 		<dd>
 			If the resolution is successful, this MUST be a <a>DID document</a> that
 			is capable of being represented in one of the conformant
-			<a data-cite="did-core#representations">representations</a> of the [[[DID]]] specification.
-			The value of <code><a data-cite="did-core#dfn-id">id</a></code> in the resolved <a>DID document</a> MUST
-			match the <a>DID</a> that was resolved. If the resolution is unsuccessful, this
-			value MUST be empty.
+			<a data-cite="did#representations">representations</a> of the <a data-cite="did#data-model"></a> 
+			specification. The value of <code><a data-cite="did#dfn-id">id</a></code> 
+			in the resolved <a>DID document</a> MUST be string equivalent to the <a>DID</a> that was resolved.
```
jyasskin: This should probably be "equal" instead of "equivalent" since equivalence classes can be larger then 1 string.

Issue #259, opened 2025-12-17
DID URL Derefencing and Fragments: The role of a DID Resolver


swcurran:
AFAIUI DID URL fragments are to be handled **after** DID resolution, by the client -- just as they are with HTTP URLs. Thus, the act of "DID Resolution" stops at returning the resource and the handling of the fragment is up to the client.

If so, [Example 5.5.1](https://w3c.github.io/did-resolution/#example-dereferencing-to-verification-method) is incorrect.  The resolution of DID Resolution is the full DID Doc + metadata. It is not just the node of the DID Doc indicated by the fragment, as explained in the current spec text.

This impacts a number of places in the spec, generally simplifying the text.

Most importantly, we should be very clear on what the concrete output of DID Resolution is expected to be -- the resource and metadata.  This simplifies creating DID Resolvers and clients that call DID Resolvers.

PR #260, opened 2025-12-18
Add DID URL Path Handling

swcurran:
This PR is a second attempt to add DID URL Path handling to the DID Resolution spec. This versions leaves the overall DID URL Dereferencing alone, and replaces the path handling component with the algorithm discussed last week. In addition to changing the previous path handling algorithm, the PR:

- Adds a “Service Types” section below “Parameters” to provide a space in the spec for formally specifying service types, and within that, the definition/algorithm of the “pathService” service type. We may decide to add other service types to this section.  For example, I’d love to see the “whoisService” added. Like the “Parameters” — DID Resolvers are generally expected to support the service types defined here.
- Drops from the path handling section the “and/or query parameters” in the introduction, and moves that up to the previous section, generalizing its introduction to cover query parameters in general, and then leaving in the specific focus on the service, service-type and relativeRef query parameters and how they are used.

***
<a href="https://pr-preview.s3.amazonaws.com/swcurran/did-resolution/pull/260.html" title="Last updated on Mar 31, 2026, 8:35 PM UTC (ea366b5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-resolution/260/5d29cbe...swcurran:ea366b5.html" title="Last updated on Mar 31, 2026, 8:35 PM UTC (ea366b5)">Diff</a>


index.html
```diff
@@ -456,6 +456,53 @@ <h2>DID Parameters</h2>
...
+		service being described by the service endpoint. Such a specification
+		may define the semantics of the service, the protocol(s) to be used to
+		interact with the service, the attributes in <code>service</code>
+		object. Such specifications MAY be defined here, within a DID Method
+		Specification, or as a DID Extensions. DID Resolvers SHOULD support the
+		service <code>type</code> values defined here, plus any others defined
+		elsewhere.
+	</p>
+	<section id="pathservice-service-type">
+		<h3><code>pathService</code> Service Type</h3>
```
```suggestion
		<h3><code>PathService</code> Service Type</h3>
```

I believe is the convention

index.html
```diff
@@ -1291,11 +1338,29 @@ <h2>Dereferencing the Resource</h2>
...
+				href="#did-parameters">here</a>), others MAY be defined by the
+				applicable <a>DID Method</a> specification, and others MAY be
+				defined in the DID Resolution Extensions
+				[[?DID-EXTENSIONS-RESOLUTION]]. Query parameters that are not
+				supported by the <a>DID URL dereferencer</a> MUST NOT affect
+				dereferencing behavior. Supported query parameters MAY
+				influence service selection, resource construction, or the form
+				of the dereferencing result.</p>
+				
+				The following process defines the processing
```
```suggestion
				The following algorithm defines the processing
```

index.html
```diff
@@ -1390,33 +1455,77 @@ <h2>Dereferencing the Resource</h2>
...
+						case-sensitive string prefix of the extracted path.
+						Among all matching services, select the one with the
+						longest matching prefix. That service MUST be used for
+						path resolution.
+						<ol class="algorithm">
+							<li>If two or more <code>services</code> have the
+							same longest matching prefix, the resolver MUST fail
+							resolution with an appropriate error.</li>
+							<li>If no matching service is found, the resolver
+							MUST return an error indicating that no applicable
```
wip-abramson: Again, lets define an error for this in https://w3c.github.io/did-resolution/#errors

index.html
```diff
@@ -1390,33 +1455,77 @@ <h2>Dereferencing the Resource</h2>
...
+						MUST be defined by <code>type</code> of the selected
+						<code>service</code>. A service type
+						specification MAY appear in the <a
+						href="#did-service-types"></a> section of this document,
+						in the applicable <a>DID method</a>
+						specification, or in the DID Resolution Extensions
+						[[?DID-EXTENSIONS-RESOLUTION]].
+						<ol class="algorithm">
+							<li>
+								The <code>pathService</code> type is (currently) the
```
```suggestion
								The <code>PathService</code> type is (currently) the
```

swcurran:
From @peacekeeper [here](https://github.com/w3c/did-resolution/pull/260#issuecomment-3670953488)
> As discussed before, I disagree with this. When it comes to service endpoints, the result of the DID URL Deferencing process should be a service endpoint URI (or list of service endpoint URIs), not a resource. The client can then choose to start a second dereferencing process to get the resource at that URI.

An interesting idea and I like the separation of layers.  I had thought the conclusion of the presentation I gave at the WG (and separately with other members) that "DID Resolution" would go further and return the referenced resource, but evidently that discussion was missed. I would add that despite my pretty long history with DIDs, I had never known until working on the previous iteration of this PR that an outcome of "DID Resolution" could be a list of URIs that are the service endpoints values of services.  Assuming that is a practical use case, then the DID Resolution outcome of a path being a URL (redirect) to a resource (but not following the redirect) seems reasonable.  Obviously, this is a minor change to the PR, but agree it has a pretty big impact on the thinking.  Thoughts from others on this?

This solidifies for me the discussion in #259 and in the WG call today that we have to be very clear on when the "DID Resolution" operation should end. In this "Path" case -- the operation ends at generating the resource URL that is returned to the client for action.  In the fragments case -- the operation ends at the DIDDoc passed back to the client and they handle the fragment.  The DID Resolution spec should only provide guidance over what the client might do.

It does leave open the question of what to do with a DID URL that has a path and a fragment.  I think in that case, the fragment is added to the end of the generated resource URL.

To summarize, DID Resolution MUST return to a client one of the following:

- a DIDDoc
- a list of serviceEndpoint URLs (via the `service` or `service-type` query parameters)
- a single URL for a resource (via a path). Question -- should this be folded into the "list" case?
  - I guess "relativeRef" also fits into this use case?
- an error.

Is that correct?

swcurran:
> I think this is much better than the previous PR #221, but to me it's still unclear what happens when there is both a path and a query string, e.g.
> 
> ```
> did:example:123456789abcdefghi/path/to/file.txt?versionTime=2025-10-01T12:00:00Z&transform=someValue
> ```
> 
> Are the parameters "versionTime" and "transform" passed to the DID Resolution algorithm, or are they appended to the service endpoint URI?

My attempt to address this is the following:

- There are parameters that are defined either in this spec/DID Method/DID extensions that are processed while resolving the DIDDoc -- versionTime, versionId are in the DID resolution spec.  Those would be removed in later steps in the processing.  - Likewise, there **could** be parameters that are defined in the same places that are to be processed during path handling (at this time, there are none).  Those too would be stripped in later steps.
- Any remaining parameters (unrecognized by the DID resolver) would be added to the URL for the resource returned to the client.  As I mentioned in my presentation last week, I'm less sure about this and would love to hear thoughts from others.

So in your example, versionTime would (obvs :-) ) be processed in resolving the DID Doc, and the undefined "transform" would be included in the generated URL returned to the client.

index.html
```diff
@@ -456,6 +456,53 @@ <h2>DID Parameters</h2>
...
+	<h2>DID Service Types</h2>
+
+	<p>
+		DID Documents MAY contain <code>services</code> each of which MUST have
+		a <code>type</code>. The <code>type</code> is to define the kind of
+		service being described by the service endpoint. Such a specification
+		may define the semantics of the service, the protocol(s) to be used to
+		interact with the service, the attributes in <code>service</code>
+		object. Such specifications MAY be defined here, within a DID Method
+		Specification, or as a DID Extensions. DID Resolvers SHOULD support the
```
```suggestion
		Specification, or as a DID Extension. DID Resolvers SHOULD support the
```

index.html
```diff
@@ -456,6 +456,53 @@ <h2>DID Parameters</h2>
 
 </section>
 
+<section id="did-service-types">
+	<h2>DID Service Types</h2>
+
+	<p>
+		DID Documents MAY contain <code>services</code> each of which MUST have
```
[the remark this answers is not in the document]
TallTed:
```suggestion
		DID Documents MAY contain `services` each of which MUST have
```

index.html
```diff
@@ -456,6 +456,53 @@ <h2>DID Parameters</h2>
...
+		a <code>type</code>. The <code>type</code> is to define the kind of
+		service being described by the service endpoint. Such a specification
+		may define the semantics of the service, the protocol(s) to be used to
+		interact with the service, the attributes in <code>service</code>
+		object. Such specifications MAY be defined here, within a DID Method
```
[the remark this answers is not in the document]
TallTed:
```suggestion
		object. Such specifications MAY be defined in this specification, within a DID Method
```

index.html
```diff
@@ -456,6 +456,53 @@ <h2>DID Parameters</h2>
...
+	</p>
+	<section id="pathservice-service-type">
+		<h3><code>pathService</code> Service Type</h3>
+
+		<p>The <code>pathService</code> is a service type used to generate the
+		URL for a resource that can be resolved based on the DID URL path.</p>
+		<p>The service MUST have the following attributes:
+			<ul>
+				<li><code>type</code>: MUST be <code>pathService</code></li>
+				<li><code>serviceEndpoint</code>: MUST be a URL that MUST NOT have a fragment component</li>
```
```suggestion
		<p>The `pathService` is a service type used to generate the
		URL for a resource that can be resolved based on the DID URL path.</p>
		<p>The service MUST have the following attributes:
			<ul>
				<li>`type`: MUST be `pathService`</li>
				<li>`serviceEndpoint`: MUST be a URL that MUST NOT have a fragment component</li>
```

index.html
```diff
@@ -456,6 +456,53 @@ <h2>DID Parameters</h2>
...
+				<li><code>prefix</code>: MUST be a valid path component of a URL</li>
```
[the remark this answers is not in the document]
TallTed:
```suggestion
				<li>`basePath`: MUST be a valid path component of a URL</li>
```

index.html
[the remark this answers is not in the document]
swcurran: The goal is for them to be defined in this section of this specification, so made that clearer.

index.html
```diff
@@ -456,6 +456,53 @@ <h2>DID Parameters</h2>
...
+
+	<p>
+		DID Documents MAY contain <code>services</code> each of which MUST have
+		a <code>type</code>. The <code>type</code> is to define the kind of
+		service being described by the service endpoint. Such a specification
+		may define the semantics of the service, the protocol(s) to be used to
+		interact with the service, the attributes in <code>service</code>
+		object. Such specifications MAY be defined here, within a DID Method
+		Specification, or as a DID Extensions. DID Resolvers SHOULD support the
+		service <code>type</code> values defined here, plus any others defined
```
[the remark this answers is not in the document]
swcurran: Fixed.

index.html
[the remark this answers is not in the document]
swcurran: I've changed all references to "pathService" to "PathService" and the attribute "prefix" to "basePath".  There are still a couple of uses of prefix in the text -- talking about the prefix part of the DID URL path that is matched to the "basePath" value.

swcurran:
From the discussion in the DID Working Group Meeting 20260108, the following captures what I think is the (very) rough consensus, and a potential path for going forward -- with questions for further discussion.

### What is returned when processing a DID URL with a path

I think we have agreed that DID Resolution when processing a DID URL Path stops at generating a URL for the resource indicated by the path -- a single value.  A question I still have is should this be a new case of what a DID resolver returns, or should it be an array as per what is returned in [Section 5.4.1](https://w3c.github.io/did-resolution/#dereferencing-algorithm-resource) sub-bullet 8.4 -- returning an array of service endpoints.

### Handling Query Parameters That Apply to the Returned URL

While this is not likely to be a particularly common use case, @peacekeeper is correct in his [comment above](https://github.com/w3c/did-resolution/pull/260#issuecomment-3682932057) that we need to avoid confusion between the DID URL components that are to be processed at DID Resolution time and those that are to be applied to the resource URL

I had proposed that "known DID Resolution query parameters" be consumed by the DID resolution process and removed, and any remaining ones apply to the resulting Resource URL, but that doesn't feel very clean, as I mentioned in my presentation at a prior WG meeting. @peacekeeper's comment proposes encoding the DID Resolution URL query parameters into a special query parameter and have all others added to the Resource URL. However, as noted in the DID Working Group call 20260108, this breaks the existing DID Model and implementations that are passing in DID Resolution Query Parameters.

Instead, the proposal was made on the call that all DID Resolution Query Parameters just be passed as they are today (e.g. `?versionId=12345`) and that we add a new DID Resolution Query Parameter (proposed name: `resourceParams`) that contains any URL components, properly encoded, that are to be added to the result URL.

So @peacekeeper's example -- `if I wanted to append /resume.pdf?pdfVersion=1234 to a service endpoint, and I wanted to use the DID Parameter versionId=12345` becomes:

- `did:example:123/resume.pdf?versionId=12345&resourceParams=%3FpdfVersion%3D1234

In this model, we are introducing into the DID Resolution spec as a "first class" query parameter `resourceParams` (or some other name) that is processed as part of the DID resolution process. This query parameter is added to the existing `versionId`, `versionTime`, `service`, `serviceType` and `relativeRef` parameters that are already specified in the DID Resolution spec.

**Very in the weeds here:** 

Note that I used "URL Components" vs. just query parameters for what is encoded in `resourceParams`, as there could be cases where someone might want to add a path to the Resource URL, and that would be allowed.  Weird, but allowed.  Likewise, since fragments are essentially ignored by the DID resolution process (they are a client concern applied after get the resolution result), if the Resource URL should have a fragment it could also be part of `resourceParams`.

### DID Method Path Handling

We have not yet discussed in a meeting if DID Method specific handling of a DID URL path should be allowed. It is in the PR for now as per the WG, but should it be removed, as requested?

peacekeeper:
> `did:example:123/resume.pdf?versionId=12345&resourceParams=%3FpdfVersion%3D1234`

I think this still "looks" a bit unclean, but I agree that this would solve my concerns about confusion and ambiguity about which parameters are used in which step.

The reason why it looks unclean is because the "versionId" parameter would affect the part before the path, which developers who are used to HTTP(S) URIs might find surprising.

But from a purely technical standpoint I think this could work.

> if DID Method specific handling of a DID URL path should be allowed

I heard in a recent call that the logic could be as follows:
- If PathService found in DID document -> apply this new, method-independent path dereferencing logic.
- Otherwise -> DID method is responsible for dereferencing the path, just like now (for backwards compatibility).

swcurran:
Interesting idea and in reading the about `relativeRef`, it seems like it would work to extend it. Nice!

- If combined with `service` or `serviceType` query parameters, the decoded `relativeRef` value is appended to the `serviceEndpoint` values of the selected service(s).
- If combined with a DID URL Path, the decoded `relativeRef` value is appended to the calculated URL.
- A `relativeRef` value must be URL encoded and may include a path and/or query parameters and/or a fragment.

The latter addition is not mentioned in the current definition, but is safe enough for how it used today.

index.html
```diff
@@ -1291,11 +1343,27 @@ <h2>Dereferencing the Resource</h2>
 					<pre class="example nohighlight">did:example:1234?hl=zQmWvQxTqbG2Z9HPJgG57jjwR154cKhbtJenbyYTWkjgF3e</pre>
 					<p class="issue">TODO: Specify the algorithm for processing the `hl` DID parameter.</p>
 				</li>
-				<li>If the <var>input <a>DID URL</a></var> contains the
-				<a href="https://www.w3.org/TR/did-core/#did-parameters">
-					DID parameter</a> <code>service</code> and/or the <a href="https://www.w3.org/TR/did-core/#did-parameters">
-					DID parameter</a> <code>serviceType</code>, and optionally the
-					<a data-cite="did-core#did-parameters">DID parameter</a> <code>relativeRef</code>:
+				<li>If the <var>input <a>DID URL</a></var> contains one or more
+				query parameters a <a>DID URL dereferencer</a> MUST examine all
```
```suggestion
				query parameters, a <a>DID URL dereferencer</a> MUST examine all
```

dmitrizagidulin (approved):

index.html
```diff
@@ -1291,11 +1343,27 @@ <h2>Dereferencing the Resource</h2>
...
+				process those that affect dereferencing behavior.
+				
+				<p>Some query parameters are defined by this specification (<a
+				href="#did-parameters">here</a>), while others can be defined by
+				the applicable <a>DID Method</a> specification, or in the DID
+				Resolution Extensions [[?DID-EXTENSIONS-RESOLUTION]]. Query
+				parameters that are not supported by the <a>DID URL
+				dereferencer</a> MUST be ignored. Supported query parameters MAY
+				influence service selection, resource construction, or the form
+				of the dereferencing result.</p>
```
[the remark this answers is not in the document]
swcurran: As discussed in the DID Working Group meeting on 2026-01-22, I've removed the change in this area from this PR. I'll open an issue on if the spec should say anything about the non-handling of parameters.

swcurran:
Per the discussion at the DID Working Group Meeting (2026-01-22), the following updates were made to the PR:

- Removed change about how to handle unrecognized parameters (opened issue #279 ).
- Update the language around how `relativeRef` is added to the DID URL Resource.
- Added a definition for the term "DID URL Resource" and referenced that definition in the spec.
- Updates based on feedback from reviewers.

The only remaining issue I see is what to do if there is DID URL with both a path and `service` or `serviceType` parameters. As currently defined the path would be silently ignored, as the handling of the parameters comes first in the DID URL Dereferencing algorithm.  I think the right answer is to add to that section an additional check "and no path component is present", so that the handling falls through to the path handling, and that `service` and `serviceType` parameters be used in the path handling to reduce the possible services tested for matching. 

Thoughts?

index.html
```diff
@@ -458,6 +467,53 @@ <h3>Datetime</h3>
...
+		be specified in this section of this specification, within a <a>DID
+		Method</a> specification, or in a DID Resolution Extension
+		[[?DID-EXTENSIONS-RESOLUTION]]. <a>DID Resolvers</a> SHOULD support the
+		service `type` values defined in this specification.
+	</p>
+	<section id="pathservice-service-type">
+		<h3>`PathService` Service Type</h3>
+
+		<p>The `PathService` is a service type used to generate the
+		URL for a resource that can be resolved based on the DID URL path.</p>
```
peacekeeper:
```suggestion
		URL for a resource that can be dereferenced based on the DID URL path.</p>
```

terms.html
```diff
@@ -121,6 +121,14 @@
   <dd>A data structure that represents the result of the <a>DID resolution</a> algorithm.
     May contain a <a>DID document</a>. See Section <a href="#did-resolution-result"></a>.</dd>
 
+  <dt><dfn data-lt="">DID resource URL</dfn></dt>
```
peacekeeper: I really don't like this term. It sounds too much like "DID URL", but in reality it will be a modified service endpoint URL that has no DIDs in it (in most cases a https://... URL).

Do we even need a new term for this? In the current algorithm, the term "selected service endpoint URL" is used.

index.html
```diff
@@ -383,10 +383,19 @@ <h2>DID Parameters</h2>
 				<code>relativeRef</code>
 			</td>
 			<td>
-				A relative <a>URI</a> reference according to <a
-					data-cite="RFC3986#section-4.2">RFC3986 Section 4.2</a> that identifies a
-				<a>resource</a> at a <a>service endpoint</a>, which is selected from a <a>DID
-				document</a> by using the <code>service</code> parameter.
+				A relative reference (path, query parameters, and/or fragment) as defined by <a
+					data-cite="RFC3986#section-4.2">RFC3986 Section 4.2</a> that is appended to either:
```
peacekeeper: As discussed, I think we should avoid terms such as "append" or "calculate" when specifying `relativeRef` or the new path handling functionality. We should just refer to the existing functionality for relative references defined in RFC3986 Section 4.2.

The reason why "append" isn't a good term is because there are some special cases such as `/./` or `/../` path segments, or cases when the both the base URI and relative reference contain query components.

index.html
```diff
@@ -458,6 +467,53 @@ <h3>Datetime</h3>
...
+		<p>During the <a href="#dereferencing-algorithm"></a> where a DID URL
+		path is present, if the selected service to be used for DID URL path
+		handling is of type `PathService`, the following MUST be
+		performed to generate the <a>DID resource URL</a>. The input to this algorithm is
+		a calculated path obtained by removing the selected `basePath` as a
+		case‑sensitive string match from the beginning of the DID URL path.
+			<ol class="algorithm">
+				<li>Obtain the `serviceEndpoint` attribute from the
+				selected service.</li>
+				<li>Append the input calculated path to the
```
peacekeeper: See previous comment, avoid the term "append", and use existing relative reference processing defined in RFC3986 Section 4.2.

peacekeeper:
It also occurs to me that the new path handling functionality could be seen as a shorthand or alternative syntax for how the `service` and `relativeRef` parameters would be used today.

Example 1:
`did:example:123/resume.pdf?versionId=12`

... would be equivalent to:
`did:example:123?service=PathService&relativeRef=/resume.pdf&versionId=12`

Example 2:
`did:example:123/resume.pdf?relativeRef=%3FpdfVersion%3D1234&versionId=12`

... would be equivalent to:
`did:example:123?service=PathService&relativeRef=/resume.pdf%3FpdfVersion%3D1234&versionId=12`

Right? Or am I missing something?

swcurran:
Updated the PR to remove the standalone DID URL Path Handling algorithm that I had added to the DID URL Dereferencing, and added the Path Handling into the existing algorithm that handles the `service`, `serviceType` and `relativeRef` query parameters.  Path and the parameters can now be used in combination with a DID URL Path (which makes sense), and the algorithm is pretty clean -- and much cleaner than having the two algorithms.

In doing that, I removed the term "DID resolution URL" that I had added, and removed the "PATH..." Error Code that I had introduced.

No outstanding questions from me on this -- I think it is completed. 🤞

msporny:
> No outstanding questions from me on this -- I think it is completed. 

I haven't had a chance to review it yet, but just wanted to send a note of appreciation -- thank you @swcurran -- I know it's been a lot of work, and the spec and community will be better off for it!

index.html
```diff
@@ -458,6 +471,51 @@ <h3>Datetime</h3>
...
+
+	<p>
+		DID Documents can contain `services`, each of which has a `type`. The
+		`type` defines the kind of service being described by the
+		`serviceEndpoint` attribute. Such a specification can define the
+		semantics of the service, the protocol(s) to be used to interact with
+		the service, and the attributes in a `service` object. Service Types can
+		be specified in this section of this specification, within a <a>DID
+		Method</a> specification, or in a DID Resolution Extension
+		[[?DID-EXTENSIONS-RESOLUTION]]. <a>DID Resolvers</a> SHOULD support the
```
```suggestion
		[[?DID-EXTENSIONS-RESOLUTION]]. <a>DID Resolvers</a> MUST support the
```

Is the reason not to do this to ensure backward compatibility with resolver implementations?

index.html
```diff
@@ -1297,134 +1355,173 @@ <h2>Dereferencing the Resource</h2>
...
+						will include a `basePath` attribute whose value is a
+						valid URL path component (<a
+						data-cite="RFC3986#section-3.3">RFC3986 Section
+						3.3</a>), and a `serviceEndpoint` attribute. At minimum,
+						<a>DID URL dereferencers</a> MUST consider `service`
+						objects with `type` values <a
+						href="#did-service-types">defined in this
+						specification</a>, plus any that are DID Method-specific
+						or DID Resolution Extensions
+						[[?DID-EXTENSIONS-RESOLUTION]].
```
wip-abramson: This normative MUST statement contradicts with the SHOULD in the Service Types section that I highlighted I believe.

Also, it is not clear where the MUST ends - is it before or after the comma?

index.html
```diff
@@ -1297,134 +1355,173 @@ <h2>Dereferencing the Resource</h2>
...
+					<li>
+						If the <a>DID URL</a> contains a path, use the `type` of
+						the one remaining `service` to produce a result URL. The
+						algorithm for assembling the result URL MUST be defined
+						by the `type` of the selected `service`. A `service`
+						object `type` specification can appear in the <a
+						href="#did-service-types"></a> section of this document,
+						in the applicable <a>DID method</a>
+						specification, or in the DID Resolution Extensions
+						[[?DID-EXTENSIONS-RESOLUTION]].
```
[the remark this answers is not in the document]
wip-abramson: The challenge I have with this approach is around interop. 
I was imagining that conformant DID Resolvers MUST support PathService. 
Does this then MUST support PathService using algorithm xxxx.
Or maybe this is not a MUST at all.

index.html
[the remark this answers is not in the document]
wip-abramson: No, don't think so. It depends what is being extended.

In this case, a new service type is being added to the DID document. The properties of the DID Document are being extended. 

In other cases, DID Resolution is being extended either by adding new parameters or new errors.

I guess there is an argument, that a new service type requires a resolution extension to define how to handle resolution (at least in this path case). But the extension originates by defining the new service type as something that goes in a DID Document. Plus all other service types have been added to the DID Document Property Extensions.

index.html
swcurran: Agree -- corrected.

wip-abramson (approved):

dmitrizagidulin (approved):

index.html
```diff
@@ -462,6 +475,79 @@ <h3>Datetime</h3>
...
+		<h3>`PathService` Service Type</h3>
+
+		<p>`PathService` is the `type` of a `service` object used to generate a
+		URL for a resource from a DID URL path. Its attribute `pathAlgorithm`
+		determines how a <a>DID URL</a> path is to be processed when its
+		`service` is selected. The processing to be used when the
+		`pathAlgorithm` value is `relative-ref-2026` (the default value) is
+		defined below. Other values can be defined for `pathAlgorithm` with
+		different processing rules within a <a>DID Method</a> specification, or in a
+		Document Property Extension [[?DID-EXTENSIONS-PROPERTIES]].
```
TallTed: Indents should always be SPACES, never TABS, because TAB settings vary with the editing software, the person doing the editing, etc. SPACES are always the same.
```suggestion
        defined below. A <a>DID Method</a> specification or a Document Property
        Extension [[?DID-EXTENSIONS-PROPERTIES]] can define other values for
        `pathAlgorithm` with their own processing rules.
```

index.html
```diff
@@ -1332,134 +1418,175 @@ <h2>Dereferencing the Resource</h2>
...
-				<a href="https://www.w3.org/TR/did-core/#did-parameters">
-					DID parameter</a> <code>service</code> and/or the <a href="https://www.w3.org/TR/did-core/#did-parameters">
-					DID parameter</a> <code>serviceType</code>, and optionally the
-					<a data-cite="did-core#did-parameters">DID parameter</a> <code>relativeRef</code>:
+
+				<li>If the <var>input <a>DID URL</a></var> contains a <a>DID path</a> and/or `service`<a href="https://www.w3.org/TR/did-core/#did-parameters">
+				DID parameter(s)</a> and/or `serviceType`<a href="https://www.w3.org/TR/did-core/#did-parameters">
+				DID parameter(s)</a> and/or a `relativeRef`<a href="https://www.w3.org/TR/did-core/#did-parameters">
+				DID parameter</a>:
+				</li>
```
TallTed: **_IF_** the rephrasing of the red (lines 1335-1339) in the green (lines 1422-1426) is correct, then I strongly suggest the revision below. My revision puts spaces around each `<a>...</a>` and not at the ends within `<a> ... </a>`. I also made an unordered `one or more` list of the `and/or` clauses, which I think is a clearer presentation.

_**HOWEVER**_, I do not think the rephrasing in the green is correct. "the DID parameter `service`" DOES NOT MEAN THE SAME as "the `service` DID parameter(s)"! The other things that are listed as possibly being found in the `input DID URL` are also different in the red than in the green.

I think that the intended meaning of this section should be confirmed (maybe the original text in red lines 1335-1339 was incorrect, and the green lines 1422-1426 are correct?), and then I will re-review for clarity in our presentation.

-----

```suggestion
                <li>If the <var>input <a>DID URL</a></var> contains one or more of the following:
                  <ul>
                    <li>a <a>DID path</a></li>
                    <li>one or more `service` <a 
                      href="https://www.w3.org/TR/did-core/#did-parameters"
                      >DID parameters</a></li>
                    <li>one or more `serviceType` <a 
                      href="https://www.w3.org/TR/did-core/#did-parameters"
                      >DID parameters</a></li>
                    <li>a `relativeRef` <a 
                      href="https://www.w3.org/TR/did-core/#did-parameters"
                      >DID parameter</a></li>
                  </ul>
                  then
				</li>
```

TallTed (changes requested):

index.html
```diff
@@ -387,10 +387,23 @@ <h2>DID Parameters</h2>
...
+				A relative reference (path, query parameters, and/or fragment)
+				as defined by <a data-cite="RFC3986#section-4.2">RFC3986 Section
+				4.2</a> combined (via the execution of the algorithm specified in
+				<a data-cite="RFC3986#section-5">RFC3986 Section 5:
+				Reference Resolution</a>) with the following `baseURI`
+				strings during <a>DID URL dereferencing</a>:
+					<ul>
+						<li>the [=DID service endpoint=](s), selected from a
+						<a>DID document</a> by using the `service` or
+						`serviceType` query parameters.</li>
```
jandrieu: I don't understand this definition. relativeRef is a parameter that is used for dereferencing, and my expectation is that it is entirely defined by RFC3986. 

The language that defines relative ref as 3986 "combined" with reference resolution creates two problems. 

First, it makes me feel that the value of relativeRef cannot be just what is in 3986, but that it's value is somehow augmented by combining it with base URI for dereferencing.  I don't think that's right. Given the pathService service type, that may be how the relative ref is applied during dereferencing, but the relative ref doesn't change because of how you dereference it. We should probably specify how PathService uses relativeRef without forcing all uses of relativeRef to assume that construction.

Second, what does it mean to combine? That seems like a misnomer. A relativeRef is not a combination of a baseURI and itself, which is how this reads: ~"A relative ref is a ref **combined** with the base." That's not quite right as the relative ref is an input to the combination algorithm during resolution & dereferencing, not the output of that combination. The output is the URL of the resource.

index.html
```diff
@@ -462,6 +475,79 @@ <h3>Datetime</h3>
...
+		DID Documents can contain `services`, each of which has a `type`. The
+		`type` defines the kind of service being described by the
+		`serviceEndpoint` attribute. Such a specification can define the
+		semantics of the service, the protocol(s) to be used to interact with
+		the service, and the attributes in a `service` object. Service Types can
+		the service, and the attributes in a `service` object. Service Types can
+		be specified in this section of this specification, within a <a>DID
+		Method</a> specification, or in a Document Property Extension
+		[[?DID-EXTENSIONS-PROPERTIES]]. <a>DID Resolvers</a> MUST support the
+		service `type` values defined in this specification.
```
jandrieu: I don't think it is viable to create a normative requirement that all resolvers MUST support PathService.

I think the right language here is that resolvers MAY support any service properties they choose, and if they choose to support the PathService property, then it MUST be implemented as defined in this section.

index.html
```diff
@@ -1332,134 +1418,175 @@ <h2>Dereferencing the Resource</h2>
...
 				</li>
-				<li>If the <var>input <a>DID URL</a></var> contains the
-				<a href="https://www.w3.org/TR/did-core/#did-parameters">
-					DID parameter</a> <code>service</code> and/or the <a href="https://www.w3.org/TR/did-core/#did-parameters">
-					DID parameter</a> <code>serviceType</code>, and optionally the
-					<a data-cite="did-core#did-parameters">DID parameter</a> <code>relativeRef</code>:
+
+				<li>If the <var>input <a>DID URL</a></var> contains a <a>DID path</a> and/or `service`<a href="https://www.w3.org/TR/did-core/#did-parameters">
+				DID parameter(s)</a> and/or `serviceType`<a href="https://www.w3.org/TR/did-core/#did-parameters">
+				DID parameter(s)</a> and/or a `relativeRef`<a href="https://www.w3.org/TR/did-core/#did-parameters">
```
jandrieu: Given that relativeRef is not used in the algorithm, I believe this is too broad.

index.html
```diff
@@ -1332,134 +1418,175 @@ <h2>Dereferencing the Resource</h2>
...
+						include only `service` objects whose `id` matches the
+						`service` parameter value(s) and/or whose `type` matches
+						the `serviceType` parameter value(s), as applicable.
+					</li> 
+					<li>
+						If the input <a>DID URL</a> contains a path component,
+						further filter the list of remaining `service` objects
+						to include only those whose `type` indicates they are
+						for <a>DID URL</a> path handling and are supported by
+						the <a>DID URL dereferencer</a>. Those `service` objects
```
jandrieu: This is too restrictive and incompatible with linkedResource.

Filtering services based on the path part of the DID URL is overly restrictive. There are services that might use the path part in other ways than are described herein.

index.html
```diff
@@ -1332,134 +1418,175 @@ <h2>Dereferencing the Resource</h2>
...
+						will include a `basePath` attribute whose value is a
```
jandrieu: Not all services that take path as input include a "basePath". That happens to be true for PathService, but we don't want to restrict other service types.

index.html
swcurran: Thanks for highlighting.  I think this is a result of the rebasing, but not sure. Either way, needs to be fixed. I do like the list -- I was thinking that was probably needed.

dmitrizagidulin (approved):

jandrieu (changes requested):
I can get behind defining a new PathService service type, but I don' think the redefinition of the relativeRef parameter makes sense. It isn't even used by the PathService algorithm.

(Also, apologies this didn't get posted sooner. I didn't realize I missed the last step to submit the review.)

swcurran:
> I can get behind defining a new PathService service type, but I don' think the redefinition of the relativeRef parameter makes sense. It isn't even used by the PathService algorithm.
> 
> (Also, apologies this didn't get posted sooner. I didn't realize I missed the last step to submit the review.)

Hey @jandrieu -- there is no intention to redefine the meaning of the relativeRef parameter, so I think it is just a clarification needed.  I (tried) to just add a line to say that relativeRef can also be applied to the result of DID URL Path handling.

Yes, it is used with DID URL Path handling. Notably, once the PathService (other path handling service) comes up with the resource URL, the `relativeRef` is used to add parameters, a fragment and perhaps even a path to the URL. For example:

- DID URL:
  - `did:web:example.com/path/to/resource.json?relativeRef=%3Ffoo%3Dbar%23myobj`
- Could produce from the PathService the resource URL:
  - `https://example.com/DIDfiles/path/to/resource.json`
- And when `relativeRef` is applied (as normal), we get a final resource URL of:
  - `https://example.com/DIDfiles/path/to/resource.json?foo=bar#myobj`

By using the `relativeRef` parameter, there is no ambiguity about what parameters might apply to DID resolution/DID URL dereferencing vs. to the resource URL.  All parameters are for DID resolution/DID URL dereferencing and if you want to add anything to the resource URL you use `relativeRef` -- which the DID URL dereferencer handles.

index.html
swcurran: As discussed in previous comments, relativeRef is used in the path handling algorithm, so I think it is applicable.

index.html
```diff
@@ -1332,134 +1418,175 @@ <h2>Dereferencing the Resource</h2>
...
+						valid URL path component (<a
+						data-cite="RFC3986#section-3.3">RFC3986 Section
+						3.3</a>), and a `serviceEndpoint` attribute. At minimum,
+						<a>DID URL dereferencers</a> MUST consider `service`
+						objects with `type` values <a
```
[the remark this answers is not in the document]
swcurran: We have the opposite suggestion from others. ¯\\\_(ツ)_/¯ :-)

My $0.02CDN is that anything defined in the DID Core and DID Resolution specs are "must support" features -- as applicable for the DID Methods your resolver supports. There should not be an "I don't know what that is" response for such features.  That is what this is intended to mean.

index.html
```diff
@@ -462,6 +475,79 @@ <h3>Datetime</h3>
...
+		the service, and the attributes in a `service` object. Service Types can
+		be specified in this section of this specification, within a <a>DID
+		Method</a> specification, or in a Document Property Extension
+		[[?DID-EXTENSIONS-PROPERTIES]]. <a>DID Resolvers</a> MUST support the
+		service `type` values defined in this specification.
+	</p>
+	<section id="pathservice-service-type">
+		<h3>`PathService` Service Type</h3>
+
+		<p>`PathService` is a `type` of a `service` object used to generate a
```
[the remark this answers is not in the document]
TallTed: -- or this --
```suggestion
		<p>`PathService` is the `type` of a `service` object used to generate a
```

index.html
```diff
@@ -462,6 +475,79 @@ <h3>Datetime</h3>
...
+<section id="did-service-types">
+	<h2>DID Service Types</h2>
+
+	<p>
+		DID Documents can contain `services`, each of which has a `type`. The
+		`type` defines the kind of service being described by the
+		`serviceEndpoint` attribute. Such a specification can define the
+		semantics of the service, the protocol(s) to be used to interact with
+		the service, and the attributes in a `service` object. Service Types can
+		the service, and the attributes in a `service` object. Service Types can
```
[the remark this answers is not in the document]
swcurran: Doh!  Thanks!

index.html
```diff
@@ -462,6 +475,79 @@ <h3>Datetime</h3>
...
+		<a data-cite="RFC3986#section-5">RFC3986 Section 5: Reference Resolution</a> as follows:
+			<ol class="algorithm">
+				<li>The <strong>base URI</strong> value is the <var>selected [=DID service endpoint=] URL</var>.</li>
+				<li>The <strong>relative reference</strong> is the path obtained by
+				removing the selected `basePath` as a case‑sensitive string match
+				from the beginning of the <a>DID URL</a> path.</li>
+				<li>The result is the URL to be returned to continue
+				the DID URL path handling algorithm.</li>
+			</ol>
+		</p>
```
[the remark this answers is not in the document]
swcurran: I think adding that would be more complicated -- and would have to be repeated through out. I think the RFC3986 Section 5 algo is fairly well known and thus sufficient. We don't want to repeat it.

index.html
```diff
@@ -462,6 +475,79 @@ <h3>Datetime</h3>
...
+		<pre class="example nohighlight" title="A DID URL with a Path and the
+		returned URL, assuming the `service` object from the example above">
+			did:example.com:123/path/to/resume.pdf
+			https://example.com/123/DIDResources/path/to/resume.pdf
+		</pre>
+	</section>
+
+</section>
```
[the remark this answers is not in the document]
swcurran: I think Joe and I cleared this up on our call.  In this part of the spec (def of PathService algo) the "relative reference" for the RFC 3986 Section 5 algorithm is not the `relativeRef` parameter. It is the DID URL path, with the `basePath` prefix removed.

jandrieu (changes requested):
Unfortunately, this is a partial review. There are a few things that still need updating.

index.html
jandrieu: The problem is this is in DID resolution and not all services that use the path are required to have a basePath.

swcurran:
@pchampin -- Per our Working Group conversation, I've removed the CSS styles that I added.  The "offending" unordered list is on line 1421 of my PR. If you are able clean it up, let me know as I would like to apply the "fix" to a couple of other places in this algorithm where the numbering is unecessary.

index.html
```diff
@@ -1454,21 +1455,23 @@ <h2>Dereferencing the Resource</h2>
...
-						<a>DID URL dereferencers</a> MUST consider `service`
-						objects with `type` values <a
-						href="#did-service-types">defined in this
-						specification</a>, plus any that are DID Method-specific
-						or Document Property Extensions
-						[[?DID-EXTENSIONS-PROPERTIES]].</p>
+                        further filter the list of remaining `service` objects
+                        to include only those whose `type` indicates they are
+                        for <a>DID URL</a> path handling and are supported by
+                        the <a>DID URL dereferencer</a>.
```
```suggestion
                        the <a>DID URL dereferencer</a>.</p>
```

index.html
```diff
@@ -462,6 +473,79 @@ <h3>Datetime</h3>
...
+			<ul>
+				<li>`type`: MUST be `PathService`</li>
+				<li>`serviceEndpoint`: MUST be a URL</li>
+				<li>`basePath`: MUST be a valid path component of a URL (as
+				specified in <a data-cite="RFC3986#section-3.3">RFC3986 Section
+				3.3</a>)</li>
+				<li>`pathAlgorithm`</li> OPTIONAL. A string value identifying
+				the algorithm used to process the DID URL path. If the
+				`pathAlgorithm` attribute is not present, its value MUST be
+				assumed to be `relative-ref-2026`.
```
```suggestion
				<li>`pathAlgorithm`: OPTIONAL. A string value identifying
				the algorithm used to process the DID URL path. If the
				`pathAlgorithm` attribute is not present, it MUST be
				treated as if set to `relative-ref-2026`.</li>
```

jandrieu:
> This is a note about the backwards compatibility of the DID URL Path handling algorithm WRT the [DID Linked Resources](https://w3c.github.io/did-resolution/) (DLR) spec. it justifies this PR on multiple levels and I really think it shows exactly why we need to merge it. For those following along, I’ll reference the version of the [DID Resolution spec with my PR](https://swcurran.github.io/did-resolution/), and I’ll focus on the algorithm in question — [Section 6.4.1 Dereferencing a Resource](https://swcurran.github.io/did-resolution/), Step 8. Happy to do this as well for any other DID Path Handling Extension.
> 
> 1. **Backwards Compatible**: Assuming there is either no PathService type services in the DID Doc, or no PathService type services that match the path, Step 8.5 would kick in and the if the resolver knows about the DLR spec would use it. Step 8.5 handles the fallback, meaning DLR implementations are unaffected — backwards compatibility is preserved.
> 2. **But if a PathService Is Added?:** @jandrieu argued in the call that an implementor might have a working DLR implementation, add a PathService to their implementation because the DID Resolution now says to, and break their working implementation. While that might be a risk, I think it would be improbable.

There's a significant miscommunication here. I'm speaking about linked resources, as listed in the document property extension registry https://www.w3.org/TR/did-extensions-properties/#linkedresource . DID linked resources is a completely different, incompatible algorithm that uses DID document metadata for describing the linked resource. It is a Draft Community Group Report of the CCG. https://w3c-ccg.github.io/did-linked-resources/

>    1. The spec does not say a PathService is required in a DIDDoc, so there is no requirement to add one.  The (revised) spec does say that a path SHOULD be handled, and that if a PathService is present here is how it must be handled. But if there is no PathService that handles a given path the backwards compatibility still works.

See my example below. Introducing a PathService can break an existing linked Resource.

>    2. A PathService cannot be added to a DIDDoc without an configuring where resources should be found — setting up `basePath` and `serviceEndpoint`. It is unlikely that an implementor would add a PathService to their DID without considering the parameters for it. Only if they use attribute settings that clash with the DLR pathing will they have a problem. If the DID Controller has gone to the effort of supporting DLRs (or some other specialized path), it’s unlikely they would blindly add a not required, deliberately conflicting PathService entry in their DID Doc.

Actually, I think it is quite likely that people will blindly add things to their DID document as service providers will simply guide users through updating their DID to support their service.

It is the possibility of "attribute settings that clash" that is the problem. Those end-users who authorize software to add service endpoints will not be able to tell which edit breaks things. In fact, most software implementers will also not know how to resolve this conflict. Other properties in the DID document lack this kind of interference risks. I can add a new verification method or service without invalidating existing methods or services. If you enable a world where it is possible to have conflicting properties, the security analysis starts with accepting that possibility is an inevitability for some, unknown number of users.

>    3. A mitigation is that we could add a note to the spec that existing implementations that already supports an extension that does Path Handling, care should be taken in adding a PathService to prevent a conflict with any existing path handling.  However, doing that waters down this important PR.

As Manu suggested, this wouldn't address my concerns.

> 3. **DLR was created because there is not standardized DID URL Path Handling**: Cheqd had a problem — how to put “resources” (files) onchain, and resolve them. They solved it by creating and implementing DLR. AFAIK, the Cheqd use of DLR translates the DID into URL that is resolved to retrieve the resource (e.g. https://resolver.cheqd.net/1.0/identifiers/did:cheqd:testnet:z6jKUJA5.../resources/a20aa56a-...). Using the DID URL Path Handling as we’ve defined it in this PR in place, they can just use PathService as is, with appropriate configuration of the basePath and serviceEndpoint attributes. With that, they can also include support for other types of resources (e.g. off chain) without impacting their implementation.

I won't comment on the shortcomings relative to DLR, as I'm not an expert on the how and why they designed their extension the way they did.

> 4. **The DLR Spec is “challenging”**: The DLR spec attempts to simultaneously satisfy Cheqd's specific on-chain resources requirements and be DID Method-neutral, which results in unnecessary complexity. I don’t think we should hold back the DID URL Path Handling PR based on the DLR work. Further, the DLR spec exemplifies exactly why we need the DID URL path handling PR merged into the spec. If DID URL Path handling had been in place earlier, DLR simply wouldn’t have been needed. 

Frankly, I doubt that. We had a path handling capability in the ecosystem with linkedResources, and we could have updated that to address @swcurran's needs and cheqd also could have built on it. Instead, people like to reinvent the wheel, and now we have three completely incompatible approaches to handling DID URL paths.

> Specific issues I see in the DLR spec:
>    
>    1. In the DLR spec, the DID URL Derefrencer is expected to retrieve and return the resource, which is beyond what a DID URL Dereferencer is expected to do, per the spec.  

That is incorrect; likely its a continuation of the unfortunately conflation of dereferencing and resolution. A DID URL dereferencer is absolutely expected to return the resource. That is what dereferencing *is*. 

**from the spec**
> The [DID URL dereferencing](https://www.w3.org/TR/did-core/#dfn-did-url-dereferencing) function dereferences a [DID URL](https://www.w3.org/TR/did-core/#dfn-did-urls) into a resource. with the following abstract form: 
<pre>
dereference(didUrl, dereferenceOptions) →
   « dereferencingMetadata, contentStream, contentMetadata »
</pre>


It is **resolution** that does not return the resource. A conformant resolver will (once we update the text) only ever return a DID document and related metadata. A conformant dereferencer will return the resource content-stream itself, including any requested transformations and related metadata.

> This the same mistake that I originally did in the first try at the DID URL Path Handling PR — had the DID URL Dereferencer resolving URLs and retrieving resources. As such, the DLR needs to be updated to reflect the boundary of DID URL Dereferencing.

By my read, the DLR spec has it correct with dereferencing.

>    2. There are multiple ways to reference a DID-linked resource in the DLR spec — a path, parameter or fragment. Its not clear why there needs to be multiple ways, or what to do if all are used at once — especially if they reference different resources.

Agreed. Multiple options are confusing. Although I can appreciate that a service query parameter SHOULD enable any one of multiple different services to be selected, and each of those may choose to interpret the path in any way they prefer. For example, did:ex:abc/images/hat.png?service=MyPathService, should be able to  unambiguously pick a specific service without conflict.

>    3. The assumption is that whatever is done is entirely up to the DID Method.  That might be good for a resource where there is no URL, but overly complex where there is a URL to resolve. I would further venture that it would be easier to independently define a URL for an on-chain resource that could be used with DIDs and other tools rather than to implement such a thing within custom DID Method just to support DLR.

I'd wager against that venture. Many DIDs *are* URLs for on-chain resources. This mythical URL for an on-chain resource that could be used with DIDs *is* a DID URL. You literally can't get to onchain resources via current http(s) URLs (nor any other URL schemes).

Perhaps more vital: the assumption is not that things are method-dependent. It's worse than that. It's ANY property added to the DID document or DID document metadata, as well as any DID method specific extensions. `linkedResource` is not defined for just did:cosmos DIDs. Its defined for ANY DID with that property in its DID document.

On a different note: you might find some value in the partial revision I did at https://jandrieu.github.io/did-resolution/#did-service-types .  Mostly, I just updated the definitions section for clarity. By the time I got to rewriting the dereferencing algorithm, I ran into conceptual problems that need more attention.

The backwards compatibility problem is that users will unwittingly add multiple, incompatible properties and not understand why their DID doesn't work.

To wit: Start with a DID URL `did:example:abc/image.png` whose DID document uses the linkedResource definition from the extensions registry https://www.w3.org/TR/did-extensions-properties/#linkedresource, to find the image.png file and verify its hash.

Then, on the advice of some service provider, the user adds a PathService property with a base path of "/".

With @swcurran's current proposal, not only is the image.png served from a different service provider (the one in the PathService object), it is no longer integrity checked against the hash. This is a significant security concern and would allow anyone to inject their preferred path service handler for all resources in the DID document and user's won't understand it is happening.

To Manu's point, simply leaving a note for implementers would not address my concerns.

I can see a few ways forward.

1. Do nothing in DID Resolution. Continue to allow for did-method specific and property-extension specific variations on path handling. PathService can be added as an extension, making it a peer of linkedResources and didLinkedResources and any other property that might be added in the future. This would mean creating a standalone Relative Ref specification (and not something included in the DID resolution spec) to ensure parity with the other approaches. This is the lowest hanging fruit that supports PathService.

2. Find a conflict resolution algorithm that allows for a DID document to specifically state the one and only one path algorithm to use, such that linkedResources and didLinkedresources are equal peers with Relative Ref 2026. This includes bringing in the spec definitions from those outside publications into DID resolution. It would also include normative language clarifying how path handling algorithms MUST identify themselves as of a shared type, e.g., `"type":"PathService"` and that document updates MUST check to see there is one and only one PathService defined.

I believe this is essentially what @swcurran was attempting to set up. However, the proposed PR fails in several regards.

-   a. It provides multiple ways for a DID URL creator to specify the use of a particular path algorithm. The resulting "greedy" search over the path to match against arbitrary number of possible service objects creates multiplicity is too confusing for predictable use. 
-   b. In particular, the greedy string matching algorithm means that even if trying to following the fallback approach, of using, e.g., both a linkedResource and a pathService property means that unexpected interactions can occur based on misunderstanding the search filter.
-   c. The PR provides almost none of the features of linked resources, which I enumerate in more detail below. It also does not address at least some of the features of did linked resources, but I'll let Alex Tweeddale chime in on what's important to them.

3. Develop a unified path handling mechanism that supports all of the features of pathService, linked Resources, and did linked Resources, such that everyone can move to this future foundation. Unfortunately, this might seem cleanest, but it doesn't address the possibility that future path-handling extensions may be created.

Currently, path service does not provide for many of the features in link resources.

To wit:

```jsonc
{
 "linkedResource": [{

  "path"(optional): // IID Resource path for this resource in the asset namespace, e.g., /myResource.png

   "id"(optional): // IID Reference ID for this resource, e.g., #myResource.png

   "rel"(optional): // the relationship of this resource to the IID Asset

   "type": "nft:ResourceDescriptor", // The JSON-LD type of this resource

  "proof": [ // an array of proofs, any of which may be used
   {
    "type"(optional): "hash" | "hashgraph" | "hashset" // the form of proof used to verify the resource as authentic
    "stage"(optional): "raw" | "compressed" | "encrypted" | "encoded" // the 
    "value"(optional): hash | hashgraph | hashset, // the actual proof for this particular resource
   }
  ],
  "resourceFormat": media type, // the IANA media-type of the linked resource

  "compression"(optional): "gzip" | "none", // the compression (performed on the raw resource) of an inline resource

  "encryption": [open ended
   for arbitrary extensibility
  ], // the encryption technique applied after compression and before encoding

  "encoding"(optional): "native" | "multibase" | "string", // the encoding (after compression and encryption) of an inline resource; "native" means the inline resource is native JSON-LD with neither compression nor encryption

  "endpoint"(optional): url, // a URL at which this resource can be retrieved before decrypting and decrompressing

  "resource"(optional): "..."
  a representation of the resource
  for inline
  distribution,
  encoded as per "encoding"
    }]
}

```

None of these properties (including the **optional** endpoint) are supported in PathService.

Perhaps more importantly, linkedResource is not a service type because we do not require an endpoint, as all services do. We literally couldn't use services to achieve its goals. Specifically, in linkedResources, you can do the following, which can't be done with PathService:
1. embed the resource in the did document, including arbitrary representations, encryption, compression, and encoding
2. provide only a hash of the resource without revealing an endpoint to get it
3. provide only a hashgraph of multiple resources without revealing any endpoint nor the number of resources included.
4. define the type of the linked resource for semantic understanding of what is to be expected. This is essentially a JSON-LD alternative to the pathAlgorithm property
5. define the relationship of the linked resource to the DID subject, e.g., AuthorName or DisplayImage
6. define a resource that uses something other than a URL for dereferencing. This is particularly valuable for resources, like ordinals, that are stored on-chain.
7. provide a way to refer to the linkedResource data object via JSON-LD ID, so that a #MyID is a reference to the JSON and NOT a reference to an external resource. That is, did:example:abc#myPhoto can be used in RDF statements about that reference in the DID document while did:example:abc/myPhoto indicates an external resource.

Frankly, it would be easier to add the `Relative Ref 2026` algorithm to the linkedResource framework than the other way around.

What's not acceptable is to ignore the previous two iterations of linkedResource and didLinkedResources and claim we solve path ambiguity by forcing everyone to use yet a third, novel approach. Especially one that breaks existing production use of DID URLs.

swcurran:
Complete disaster in trying to resolve the conflicts -- `git rebase` was not a help.  AFAICT the most recent set of changes I added are gone.

¯\\\_(ツ)_/¯

I'll have to see if I can find the changes and reapply against the latest main.  Closing this PR.

PR #271, opened 2026-01-13
Rename Read to Resolve and clarify introduction

### LINKED ISSUE #230: Unclear definition of "Read" operation for DID methods
> This refers to 'the "Read" operation of the applicable [DID method](https://www.w3.org/TR/did-resolution/#dfn-did-method-s) as described in [Method Operations](https://www.w3.org/TR/did-core/#method-operations)', but https://www.w3.org/TR/did-core/#method-operations doesn't mention the word "Read". What's the signature of that operation?

From https://github.com/w3ctag/design-reviews/issues/1157#issuecomment-3439062470

DID v1.1 spec states:

> A [DID method](https://www.w3.org/TR/did-1.1/#dfn-did-methods) specification MUST specify how a [DID resolver](https://www.w3.org/TR/did-1.1/#dfn-did-resolvers) uses a [DID](https://www.w3.org/TR/did-1.1/#dfn-decentralized-identifiers) to resolve a [DID document](https://www.w3.org/TR/did-1.1/#dfn-did-documents), including how the [DID resolver](https://www.w3.org/TR/did-1.1/#dfn-did-resolvers) can verify the authenticity of the response.

So perhaps we change DID Resolution to use the "Resolve" operation. 

Also this:

> What's the signature of that operation?

Makes me wonder if we should define resolve(did, resolutionOptions) as the interface, making it clearer that resolutionOptions get passed through to DID methods after generic processing by the DID resolver.

Also, how do DID Method specific errors get surfaced by the DID resolver?

wip-abramson:
This is an attempt to resolve #230 - or at least show one way that we might address #230.

Reviews and improvements welcome. Especially for the introduction.

***
<a href="https://pr-preview.s3.amazonaws.com/wip-abramson/did-resolution/pull/271.html" title="Last updated on Feb 2, 2026, 2:22 PM UTC (63bb0a2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-resolution/271/b78b770...wip-abramson:63bb0a2.html" title="Last updated on Feb 2, 2026, 2:22 PM UTC (63bb0a2)">Diff</a>


wip-abramson:
Thanks @peacekeeper, I understand where you are coming from. I put this together to show what the changes discussed in #230 would look like. 

I agree there is a distinction between the 1) and 2) that you point out, but I do think that they are both aspects of resolve.

1) Is the standard way that you invoke a resolver.
2) Is the DID method specific way you execute a resolve operation.

> I feel a bit that this PR conflates two things that are not the same. With this PR, we are defining a "resolve()" algorithm in this specification, while also saying that each DID method must define how to "resolve" a DID. Isn't that a contradiction?

DID methods must define how you resolve a DID makes more sense to me than: "DID methods must define how you read a DID."

The DID Resolution spec is defining a common interface to invoke a resolver independent of any DID methods specific resolve operation. It provides an interoperable mechanism to execute resolution requests for any DID. The DID resolution specification does not really define resolution, the "resolve()" algorithm is a thin wrapper that can be placed around any DID method specific resolution operation/algorithm.

wip-abramson:
@jandrieu since I know this is an issue you feel strongly about I would appreciate your review on this PR

dmitrizagidulin:
(Should be ready to merge, would be great to have one more review by Joe).

Issue #226, opened 2025-10-24
Inconsistent use of resolution?


wip-abramson:
> The use of "resolution" here seems inconsistent with other uses. In particular, I don't think it's "the process of determining an access mechanism and the appropriate parameters necessary to dereference a URI" from https://www.rfc-editor.org/rfc/rfc3986.html#section-1.2.2, and it's not "the process of resolving a URI reference within a context that allows relative references so that the result is a string matching the <URI> syntax rule" from https://www.rfc-editor.org/rfc/rfc3986.html#section-5 (which is also used in https://www.w3.org/TR/did-1.1/#relative-did-urls). Is it the right term? Perhaps you're just defining how to "dereference" both DIDs and DID URLs?

https://github.com/w3ctag/design-reviews/issues/1157

wip-abramson:
We should do a full review of the spec to check we are being consistent with our use of dereference and resolve throughout.

w3cbot:
This was discussed during the [#did meeting on 30 October 2025](https://www.w3.org/2025/10/30-did-minutes.html#e493).

<details><summary><i>View the transcript</i></summary>

<h4><a href="https://github.com/w3c/did-resolution/issues/226" rel="noopener noreferrer">w3c/<wbr>did-resolution#226</a></h4>
<p><cite>Wip:</cite> Should we be using resolution at all?  That's what he is questioning.</p>
<p><cite>&lt;Zakim&gt;</cite> manu, you wanted to note it's the first thing, isn't it?</p>
<p><cite>markus_sabadello:</cite> We've always been using resolution. The DID can also be dereferenced. Consistent with RFC3986 in my mind.  This does fit with how we are using it. I wouldn't change anything.</p>
<p>URI "resolution" is the process of determining an access mechanism and the appropriate parameters  necessary to dereference a URI; this resolution may require several</p>
<p>   iterations.  To use that access mechanism to perform an action on the</p>
<p>   URI's resource is to "dereference" the URI.</p>
<p><cite>manu:</cite> Agree with Markus.  Looking at RFC3986 -- pasting...<br>
<span>… what we are doing is textbook definition. Perhaps he is worried about a sentence and how we are using it.  Perhaps he is saying it's "dereferencing" vs. "resolution".</span></p>
<p><cite>&lt;Zakim&gt;</cite> manu, you wanted to note we need to do a complete review</p>
[...]

PR #315, opened 2026-03-31
major rewrite to address resolution/dereferencing confusion. take 2

jandrieu:
Replaces PR #311

I believe I have responded to all of the comments and suggestions in that PR.

Hopefully this one is easier to compare.

FWIW, I *would* move  the dereferencing section ahead of the resolution section, as per my comment. However, since people complained about not being able to tell where the changes were, I reverted that edit. Otherwise, both sections would look like a complete rewrite.

***
<a href="https://pr-preview.s3.amazonaws.com/LegReq/did-resolution/pull/315.html" title="Last updated on Apr 12, 2026, 8:44 PM UTC (ad17e20)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-resolution/315/ea94ecc...LegReq:ad17e20.html" title="Last updated on Apr 12, 2026, 8:44 PM UTC (ad17e20)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -202,7 +202,8 @@ cryptographic public keys. This specification covers the algorithms and guidelin
 		and associated metadata that defines the
 		cryptographic material and service endpoints for securely
 		interacting with the associated resource.
-		The client dereferencing the DID URL uses the result returned from resolution to retrieve the actual resource and apply it to the
+		The client dereferencing the DID URL uses the result returned from
+		resolution to retrieve the actual resource and apply it to the
 		current computational context, either by displaying the resulting
 		resource or extracting information from the DID document to
 		perform some other function, such as evaluating a given proof
--- a/index.html
+++ b/index.html
@@ -1199,7 +1199,7 @@ external resources.
 					URL to create a [=base DID URL=], e.g.,
 					`did:ex:abc#key-1` becomes `did:ex:abc` .</li>
 				<li><strong>Prepare</strong> <var><a>resolution
-							options</a></var>. See the [[resolve]] algorithm for
+							options</a></var>. See the <a href="#resolving">resolve</a> algorithm for
 					options.</li>
 			</ol>
 		</section>
--- a/index.html
+++ b/index.html
@@ -175,7 +175,7 @@
 	</p>
 	<p>
 		Client software dereferences identifiers to retrieve the value of a
-		reference and	bring it into the current computational context. Dereferencing a DID URL is what DID-enabled applications do to bring the referent of that DID URL into the current context.
+		reference and bring it into the current computational context. Dereferencing a DID URL is what DID-enabled applications do to bring the referent of that DID URL into the current context.
 	</p>
 	<p>In programming languages like <strong>c</strong> and
 		<strong>c++</strong>, dereferencing a pointer means accessing,
--- a/index.html
+++ b/index.html
@@ -205,7 +205,7 @@ cryptographic public keys. This specification covers the algorithms and guidelin
 		The client dereferencing the DID URL uses the result returned from
 		resolution to retrieve the actual resource and apply it to the
 		current computational context, either by displaying the resulting
-		resource or extracting information from the DID document to
+		resource or by extracting information from the DID document to
 		perform some other function, such as evaluating a given proof
 		against verification methods in the DID document.
 	</p>
--- a/index.html
+++ b/index.html
@@ -254,18 +254,18 @@ cryptographic public keys. This specification covers the algorithms and guidelin
 				metadata.
 			<p>
 
-	<p>With that DID document, clients may
-	<ol>
-		<li>further dereference the DID URL to retrieve a secondary
-			resource&mdash;this is necessary when a DID URL contains a
-			fragment or when a web server returns a redirect;</li>
-		<li>return the DID document to another process, </li>
-	  <li>display the final dereferenced result to the current user,
-		using the resource's media type, or</li>
-  	<li>if the final resource is a DID Document, interpret the DID
-		document to evaluate verification methods or other properties.
-	</li>
-	</ol>
+    <p>With that DID document, clients can do any combination of zero or more
+       of the following:</p>
+    <ol>
+        <li>Further dereference the DID URL to retrieve a secondary resource.
+            This is necessary when a DID URL contains a fragment or when a web
+            server returns a redirect.</li>
+        <li>Return the DID document to another process.</li>
+        <li>Display the final dereferenced result to the current user, using
+            the resource's media type.</li>
+        <li>If the final resource is a DID Document, interpret the DID
+            document to evaluate verification methods or other properties.</li>
+    </ol>
 	<p>The resolving client MAY choose any of these options
 		depending on the context in which the DID URL is used. </p>
 
--- a/index.html
+++ b/index.html
@@ -281,7 +281,7 @@
 	<code>#key-1</code> verification method from the DID document returned from passing the BASE DID URL <code>did:example:abc</code> to a resolver.
 	.
 	<pre class="example nohighlight"
-		title="A DID URL with a 'versionTime' DID parameter">
+		title="A Data Integrity proof produced by did:example:abc#key-1">
 				{
 					"proof": {
 					"type": "DataIntegrityProof",
--- a/index.html
+++ b/index.html
@@ -221,9 +221,9 @@ cryptographic public keys. This specification covers the algorithms and guidelin
 	<p>
 		DIDs can be used as identifiers for both subjects and issuers of
 		Verifiable Credentials and can anchor any number of
-		cryptographic
-		verification relationships, such as assertion, authentication, capability invocation and delegation, and key agreement.
-		encryption, and key agreement.
+		cryptographic verification relationships, such as assertion,
+		authentication, capability invocation and delegation, and key
+		agreement.
 	</p>
 	<p>
 		A discussion of Decentralized Identifier use cases can be found
--- a/index.html
+++ b/index.html
@@ -349,7 +349,8 @@ reading this specification.
 		  Use Cases
 		</h2>
 		<p>
-The DID resolution specification is intended to support a broad range of use cases including:
+The DID resolution specification is intended to support a broad range of use
+cases, including the following:
 		</p>
 		<ul>
 			<li>
--- a/index.html
+++ b/index.html
@@ -269,9 +269,9 @@ cryptographic public keys. This specification covers the algorithms and guidelin
 	<p>The resolving client MAY choose any of these options
 		depending on the context in which the DID URL is used. </p>
 
-	<p>For example, the following html snippet would cause the
+	<p>For example, the following HTML snippet would cause the
 		associated image to be shown in the browser displaying the
-		html page containing it.
+		HTML page containing it.
 	</p>
 
 	<pre class="example nohighlight"
--- a/index.html
+++ b/index.html
@@ -315,8 +315,10 @@ cryptographic public keys. This specification covers the algorithms and guidelin
 		</p>
 
 		<p>
-			A <dfn class="lint-ignore">conforming DID URL client</dfn> is any application or algorithm realized in software and/or hardware that complies with the relevant
-			normative statements in <a href="#dereferencing"></a>.
+            A <dfn class="lint-ignore">conforming DID URL client</dfn> is any
+            application or algorithm realized in software and/or hardware that
+            complies with the relevant normative statements in
+            <a href="#dereferencing"></a>.
 		</p>
 		<p class="note" title="Dereferencing is local">
 			There is no network-based DID URL client as dereferencing is
--- a/index.html
+++ b/index.html
@@ -219,7 +219,7 @@
 		DIDs can be used as identifiers for both subjects and issuers of
 		Verifiable Credentials and can anchor any number of
 		cryptographic
-		verification methods, such as attestations, authorizations,
+		verification relationships, such as assertion, authentication, capability invocation and delegation, and key agreement.
 		encryption, and key agreement.
 	</p>
 	<p>
--- a/index.html
+++ b/index.html
@@ -175,7 +175,9 @@
 	</p>
 	<p>
 		Client software dereferences identifiers to retrieve the value of a
-		reference and bring it into the current computational context. Dereferencing a DID URL is what DID-enabled applications do to bring the referent of that DID URL into the current context.
+		reference and bring it into the current computational context.
+		Dereferencing a DID URL is what a DID-enabled application does to
+		bring the referent of that DID URL into the current context.
 	</p>
 	<p>In programming languages like <strong>c</strong> and
 		<strong>c++</strong>, dereferencing a pointer means accessing,
--- a/index.html
+++ b/index.html
@@ -279,10 +279,12 @@ cryptographic public keys. This specification covers the algorithms and guidelin
 				&lt;img src="did:example:abc/image.png?service=pathService"&gt;
 				</pre>
 
-	Alternatively, a verifier working with the following [=verifiable
-	credential=] proof property would evaluate the proof using the
-	<code>#key-1</code> verification method from the DID document returned from passing the BASE DID URL <code>did:example:abc</code> to a resolver.
-	.
+    <p>
+        Alternatively, a verifier working with the following [=verifiable credential=]
+        proof property would evaluate the proof using the <code>#key-1</code>
+        verification method from the DID document returned from passing the BASE
+        DID URL <code>did:example:abc</code> to a resolver.
+    <p>
 	<pre class="example nohighlight"
 		title="A Data Integrity proof produced by did:example:abc#key-1">
 				{
--- a/index.html
+++ b/index.html
@@ -131,18 +131,14 @@
   <body data-cite="infra rfc3986 did">
     <section id='abstract'>
       <p>
-Decentralized identifier (DID) resolution is the process of obtaining
-a DID document and accompanying metadata for a specific DID. The process
-takes a DID and a set of resolution options as its input and returns a
-DID document and associated metadata about the resolved document and the
-resolution request. A resolved DID document is a set of information which
-enables cryptographically verifiable interactions with the DID subject,
-including mechanisms such as cryptographic public keys. This specification
-covers the algorithms and guidelines to be used for DID resolution and
-relies on the core DID specification,
-[[[DID]]],
-which describes the underlying DID architecture in full detail.
-      </p>
+			Decentralized identifier (DID) resolution is the process of
+			obtaining the authoritative DID document and associated metadata,
+			taking into account any query parameters that affect DID
+			document selection. The resulting DID document contains
+			information that enables cryptographically verifiable interactions
+			with the DID controller related to the DID subject, including, e.g.,
+			cryptographic public keys. This specification covers the algorithms and guidelines to be used	for DID resolution and dereferencing DID URLs, relying on the core DID specification [[[DID]]] for DID and DID URL syntax and the DID	document data format.
+			</p>
     </section>
 
     <section id='sotd'>
@@ -174,59 +170,136 @@ which describes the underlying DID architecture in full detail.
 
 <section id="introduction">
 <h1>Introduction</h1>
+	<p>
+		DID Resolution is the first step in dereferencing a DID URL.
+	</p>
+	<p>
+		Client software dereferences identifiers to retrieve the value of a
+		reference and	bring it into the current computational context. Dereferencing a DID URL is what DID-enabled applications do to bring the referent of that DID URL into the current context.
+	</p>
+	<p>In programming languages like <strong>c</strong> and
+		<strong>c++</strong>, dereferencing a pointer means accessing,
+		interpreting, or applying the value pointed to by that pointer
+		in the current computational context. On the web, dereferencing
+		a URL means retrieving a representation of
+		a resource and returning it for interactions in the local
+		context. For web pages, that means displaying the retrieved HTML
+		in the browser, with an appropriate view. For asynchronous
+		JavaScript, aka AJAX requests, it means bringing the result from
+		an HTTP request into the current JavaScript context for processing,
+		often to change the content of the page. Dereferencing a DID URL
+		means retrieving the appropriate resource referred to by the DID
+		URL and applying it in the current context.
+	</p>
+	<p>
+		Resolution acquires the
+		authoritative metadata that enables resource retrieval.
+		Resolving an HTTP URL means resolving the authority part
+		(typically a host or domain name) to get the IP address to use
+		for the HTTP request. Resolving a DID URL gets the DID document 
+		and associated metadata that defines the
+		cryptographic material and service endpoints for securely
+		interacting with the associated resource.
+		The client dereferencing the DID URL uses the result returned from resolution to retrieve the actual resource and apply it to the
+		current computational context, either by displaying the resulting
+		resource or extracting information from the DID document to
+		perform some other function, such as evaluating a given proof
+		against verification methods in the DID document.
+	</p>
+	<p>
+		DID URL resolution depends on method-specific interactions with
+		the DID method's Verifiable Data Registry. Every DID method
+		defines its own approach to retrieving DID state from a VDR. DID
+		Resolvers provide a standard interface to abstract those VDR
+		interactions away, enabling any DID-enabled application to rely
+		on a consistent API, regardless of how the method interacts with
+		a VDR.
+	</p>
+	<p>
+		DIDs can be used as identifiers for both subjects and issuers of
+		Verifiable Credentials and can anchor any number of
+		cryptographic
+		verification methods, such as attestations, authorizations,
+		encryption, and key agreement.
+	</p>
+	<p>
+		A discussion of Decentralized Identifier use cases can be found
+		in the <a href="https://www.w3.org/TR/did-use-cases/">W3C's Use
+			Cases	and	Requirements for Decentralized Identifiers</a>.
+	</p>
+	<p>This specification defines how DID URL enabled applications
+		can dereference a DID URL into the current context, including
+		how to resolve the DID URL to get the authoritative DID document
+		used for retrieving the actual resource.
+	</p>
 
-	<p>[=DID resolution=] is the process of obtaining a <a>DID document</a> for a given <a>DID</a>. This is one
-	of four required operations that can be performed on any DID ("Read"; the other ones being "Create", "Update",
-	and "Deactivate"). The details of these operations differ depending on the <a>DID method</a>.
-	Building on top of [=DID resolution=], <a>DID URL dereferencing</a> is the process of retrieving a representation
-	of a resource for a given <a>DID URL</a>. Software and/or hardware that is able to execute these processes is called
-	a <a>DID resolver</a>.
-
-	<p>This specification defines a standard interface that clients can use to execute [=DID resolution=] and <a>DID URL dereferencing</a>
-		requests, independent of any specific <a>DID method's</a> "Resolve" operation that a [=DID resolver=] supports. Additionally,
-		this specification defines requirements, algorithms including their inputs and results, architectural options, and various
-		security and privacy considerations relevant to implementing a [=DID resolver=] or <a>DID URL dereferencer</a>.</p>
-
-	<p>Note that while this specification defines some base-level functionality for DID resolution, the actual steps
+	<p class="NOTE">While this specification defines base-level functionality for DID resolution, the actual steps
 	required to communicate with a DID's <a>verifiable data registry</a> are defined by the applicable
 	<a>DID method</a> specification.</p>
 
 	<section id="implementer-overview" class="informative">
 		<h2>Implementer Overview</h2>
-		<p>
-			By invoking a <a>DID resolver</a> using the standard `resolve(did, resolutionOptions)` interface (as defined in
-			the <a href="#resolving">DID Resolution section</a>), one can obtain a <a>DID document</a> and accompanying metadata
-			(e.g., `contentType`, proof, versioning), which an application can use to validate a user's cryptographic keys,
-			service endpoints, or status.
-		<p>
-			For example, to retrieve the state of the DID `did:example:123`
-             at a time in the past, a client application could resolve the DID with the `versionTime`
-             resolution option, as in the following:<br>
-             `resolve("did:example:123", {"versionTime":"2021-05-10T17:00:00Z"})`
+			<p>When using a DID URL to interact with a resource, first
+	perform resolution, then apply that result to the relevant
+	workflow.</p>
+			<p>
+				By invoking a <a>DID resolver</a> using the standard
+				`resolve(didURL, resolutionOptions)` interface (as defined in
+				the <a href="#resolving">DID Resolution section</a>), one
+				obtains
+				the authoritative <a>DID document</a> and accompanying
+				metadata.
+			<p>
 
-			Or a client could
-			dereference a DID URL
-			 like <code>did:example:123?service=files&relativeRef=/resume.pdf</code>
-			 (see <a href="#example-dereferencing-to-service-endpoint-url">here for detailed example</a>)
-			 to fetch a user's resume stored via a service declared in the DID document.
-		</p>
-		<p>
-			Further, the specification's <a href="#dereferencing-algorithm">DID URL dereferencing algorithm</a> shows how
-			a client can follow a fragment (e.g., `#key-1`) to extract a particular verification method from
-			the <a>DID document</a> (see <a href="#example-dereferencing-to-verification-method">here for detailed example</a>).
-			In practice, implementers validate their resolver against the
-			<a href="https://github.com/w3c-ccg/did-resolution-mocha-test-suite">DID Resolution Test Suite</a> which exercises
-			normative MUSTs and error conditions (such as invalid DIDs, deactivated DIDs, unsupported methods,
-			relative URL expansion, etc.) to ensure that client applications can reliably depend on correct resolution behavior
-			across different DID methods.
-		</p>
+	<p>With that DID document, clients may
+	<ol>
+		<li>further dereference the DID URL to retrieve a secondary
+			resource&mdash;this is necessary when a DID URL contains a
+			fragment or when a web server returns a redirect;</li>
+		<li>return the DID document to another process, </li>
+	  <li>display the final dereferenced result to the current user,
+		using the resource's media type, or</li>
+  	<li>if the final resource is a DID Document, interpret the DID
+		document to evaluate verification methods or other properties.
+	</li>
+	</ol>
+	<p>The resolving client MAY choose any of these options
+		depending on the context in which the DID URL is used. </p>
+
+	<p>For example, the following html snippet would cause the
+		associated image to be shown in the browser displaying the
+		html page containing it.
+	</p>
+
+	<pre class="example nohighlight"
+		title="A DID URL used to retrieve and display an image.">
+				&lt;img src="did:example:abc/image.png?service=pathService"&gt;
+				</pre>
+
+	Alternatively, a verifier working with the following [=verifiable
+	credential=] proof property would evaluate the proof using the
+	<code>#key-1</code> verification method from the DID document returned from passing the BASE DID URL <code>did:example:abc</code> to a resolver.
+	.
+	<pre class="example nohighlight"
+		title="A DID URL with a 'versionTime' DID parameter">
+				{
+					"proof": {
+					"type": "DataIntegrityProof",
+					"cryptosuite": "eddsa-rdfc-2022",
+					"created": "2021-11-13T18:19:39Z",
+					"verificationMethod": "did:example:abc#key-1",
+					"proofPurpose": "assertionMethod",
+					"proofValue": "z58DAdFfa9SkqZMVPxAQp...jQCrfFPP2oumHKtz"
+					}
+				}
+				</pre>
 	</section>
 
 	<section id="conformance">
 
 		<p>
 			A <dfn class="lint-ignore">conforming DID resolver</dfn> is any algorithm
-			realized as software and/or hardware that complies with the relevant normative
+			realized in software and/or hardware that complies with the relevant normative
 			statements in <a href="#resolving"></a>.
 		</p>
 
@@ -237,17 +310,14 @@ which describes the underlying DID architecture in full detail.
 		</p>
 
 		<p>
-			A <dfn class="lint-ignore">conforming DID URL dereferencer</dfn> is any
-			algorithm realized as software and/or hardware that complies with the relevant
+			A <dfn class="lint-ignore">conforming DID URL client</dfn> is any application or algorithm realized in software and/or hardware that complies with the relevant
 			normative statements in <a href="#dereferencing"></a>.
 		</p>
-
-		<p>
-			A <dfn class="lint-ignore">conforming network-based DID URL dereferencer</dfn> is a
-			[=conforming DID URL dereferencer=] that additionally complies with the normative
-			statements in [[[#bindings-https]]].
+		<p class="note" title="Dereferencing is local">
+			There is no network-based DID URL client as dereferencing is
+			fundamentally a client-side function in which the software
+			retrieves (or interacts with) a remote resource.
 		</p>
-
 	</section>
 
 	<section class="informative">
@@ -256,7 +326,7 @@ which describes the underlying DID architecture in full detail.
 		</h2>
 		<p>
 This specification has three primary audiences: implementers of conformant DID methods;
-implementers of conformant DID resolvers; and implementers of systems and services
+implementers of conformant DID resolvers; and implementers of DID enabled applications 
 that wish to resolve DIDs using DID resolvers. The intended audience includes,
 but is not limited to, software architects, data modelers, application developers,
 service developers, testers, operators, and user experience (UX) specialists.
@@ -272,9 +342,7 @@ reading this specification.
 		  Use Cases
 		</h2>
 		<p>
-The DID resolution specification is intended to support a broad range of use cases
-by defining a standardized interface to resolve DIDs and dereference DID URLs
-independent of the DID method of any particular DID. These usecases include:
+The DID resolution specification is intended to support a broad range of use cases including:
 		</p>
 		<ul>
 			<li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -174,59 +170,136 @@
 
 <section id="introduction">
 <h1>Introduction</h1>
+	<p>
+		DID Resolution is the first step in dereferencing a DID URL.
+	</p>
+	<p>
+		Client software dereferences identifiers to retrieve the value of a
+		reference and	bring it into the current computational context. Dereferencing a DID URL is what DID-enabled applications do to bring the referent of that DID URL into the current context.
```
```suggestion
		reference and bring it into the current computational context. Dereferencing a DID URL is what DID-enabled applications do to bring the referent of that DID URL into the current context.
```

index.html
```diff
@@ -174,59 +170,136 @@
...
+		interactions away, enabling any DID-enabled application to rely
+		on a consistent API, regardless of how the method interacts with
+		a VDR.
+	</p>
+	<p>
+		DIDs can be used as identifiers for both subjects and issuers of
+		Verifiable Credentials and can anchor any number of
+		cryptographic
+		verification methods, such as attestations, authorizations,
+		encryption, and key agreement.
```
wip-abramson: I don't have a concrete suggestion - apologies - but I think we need to distinguish the verification methods from the things the produce (attestions, authorizations etc).

Maybe `verification method for purposes such as` Feels a bit clunky though.

index.html
```diff
@@ -174,59 +170,136 @@
...
+					"proof": {
+					"type": "DataIntegrityProof",
+					"cryptosuite": "eddsa-rdfc-2022",
+					"created": "2021-11-13T18:19:39Z",
+					"verificationMethod": "did:example:abc#key-1",
+					"proofPurpose": "assertionMethod",
+					"proofValue": "z58DAdFfa9SkqZMVPxAQp...jQCrfFPP2oumHKtz"
+					}
+				}
+				</pre>
```
wip-abramson, line 300: Not sure this example makes sense here. At least the title is wrong.

I guess this is the VC proof being evaluated.
```suggestion
	<pre class="example nohighlight"
		title="A Data Integrity proof produced by did:example:abc#key-1">
				{
					"proof": {
					"type": "DataIntegrityProof",
					"cryptosuite": "eddsa-rdfc-2022",
					"created": "2021-11-13T18:19:39Z",
					"verificationMethod": "did:example:abc#key-1",
					"proofPurpose": "assertionMethod",
					"proofValue": "z58DAdFfa9SkqZMVPxAQp...jQCrfFPP2oumHKtz"
					}
				}
				</pre>
```

index.html
```diff
@@ -467,10 +534,13 @@ <h3>Datetime</h3>
...
 	<h1>DID Resolution</h1>
+	<p class="note" title="consider reordering">
+		We should consider putting dereferencing first, which was my initial edit with this PR, but that ends up looking like a LOT more changes than there actually are, so if we accept this text, I'll add an issue to discuss putting the outter algorithm (dereferencing) first.
+		</p>
 
 	<p>
-		The <a>DID resolution</a> function resolves a <a>DID</a> into a <a>DID
-		document</a> by using the "Resolve" operation of the applicable <a>DID method</a>
+		The <a>DID resolution</a> function resolves a <a>DID</a> into the authoritative <a>DID
+		document</a> by using the "Resolve" operation defined by <a>DID method</a> in the DID itself
```
```suggestion
		document</a> by using the "Resolve" operation defined by the <a>DID method</a> identified in the DID itself 
```

index.html
```suggestion
resolve(didUrl, resolutionOptions) →
```

I think you are proposing we change this

index.html
```diff
@@ -1099,517 +1123,327 @@ <h1>DID URL Dereferencing</h1>
...
-				<figcaption style="text-align: center;">
-					Dereferencing a DID URL to a verification method.
-				</figcaption>
-			</figure>
+		<section id="service-type-selection-algorithm">
+			<h2>Service Type Selection Algorithm</h2>
+			<p>When a DID URL contains a <a>serviceType</a> parameter
+				use
+				the
+				following algorithm to determine which service should be
```
wip-abramson, line 1370:
```suggestion
				following algorithm to determine which service(s) should be
```

There are legitimate use cases for wanting to retrieve multiple services of the same type from the DID document. This text implies that only a single service should be selected.

TallTed:
> Replaces PR https://github.com/w3c/did-resolution/pull/311

Perhaps this should also be noted on #311?

peacekeeper:
If I'm reading this correctly, it seems "relativeRef" is only mentioned in the "Service Selection Algorithm", not in the "Service Type Selection Algorithm"?

Also, I wonder about this sentence:

> If the input [DID URL](https://www.w3.org/TR/did-core/#dfn-did-urls) contains the [DID parameter](https://pr-preview.s3.amazonaws.com/LegReq/did-resolution/pull/315.html#did-parameters) relativeRef and a service of type "PathService" is selected ...

Does this mean that "relativeRef" is only supported in conjunction with PathService, not with "service" and "serviceType" parameters?

peacekeeper:
Why does this remove the "expandRelativeUrls" and "verificationRelationship" options? Is this intentional or an accident?

index.html
```diff
@@ -131,18 +131,14 @@
...
-which describes the underlying DID architecture in full detail.
-      </p>
+			Decentralized identifier (DID) resolution is the process of
+			obtaining the authoritative DID document and associated metadata,
+			taking into account any query parameters that affect DID
+			document selection. The resulting DID document contains
+			information that enables cryptographically verifiable interactions
+			with the DID controller related to the DID subject, including, e.g.,
+			cryptographic public keys. This specification covers the algorithms and guidelines to be used	for DID resolution and dereferencing DID URLs, relying on the core DID specification [[[DID]]] for DID and DID URL syntax and the DID	document data format.
+			</p>
```
TallTed: Why is every line's indentation modified from the original? The first line here (`134`), for instance, is not changed except for the indentation. I'm sure that *every* subsequent word is not changed. There must be a way for your editor to help us, instead of making our lives (including your own!) more difficult!

index.html
jandrieu: Sure. This was an existing example. Your suggestion is goo.d

index.html
jandrieu: Ah... Good catch. These are verification relationships.

index.html
jandrieu: I am equally baffled. 

This PR was specifically edited ONLY those section with changes, without reformating the whole page.

index.html
```diff
@@ -174,59 +170,136 @@
...
+		Resolvers provide a standard interface to abstract those VDR
+		interactions away, enabling any DID-enabled application to rely
+		on a consistent API, regardless of how the method interacts with
+		a VDR.
+	</p>
+	<p>
+		DIDs can be used as identifiers for both subjects and issuers of
+		Verifiable Credentials and can anchor any number of
+		cryptographic
+		verification methods, such as attestations, authorizations,
```
jandrieu:
```suggestion
		verification relationships, such as assertion, authentication, capability invocation and delegation, and key agreement.
```

index.html
jandrieu: Respectfully, evaluating multiple services is fine, but dereferencing to multiple services is not a possibility.

A resolving client has the context to interpret a DID document if there are possible multiples. The result of dereferencing is not going to be returning that set, it is going to be picking one and bringing it into the current context.

Please expand on these use cases where something like <link rel="stylesheet" href="did:example:abs/styles.css"> returns multiple resources? Which one is brought into the renderer?

jandrieu:
> If I'm reading this correctly, it seems "relativeRef" is only mentioned in the "Service Selection Algorithm", not in the "Service Type Selection Algorithm"?
> 
> Also, I wonder about this sentence:
> 
> > If the input [DID URL](https://www.w3.org/TR/did-core/#dfn-did-urls) contains the [DID parameter](https://pr-preview.s3.amazonaws.com/LegReq/did-resolution/pull/315.html#did-parameters) relativeRef and a service of type "PathService" is selected ...
> 
> Does this mean that "relativeRef" is only supported in conjunction with PathService, not with "service" and "serviceType" parameters?

That's an oversight. We should probably apply that in both cases.

index.html
jandrieu: You've got me. I explicitly avoided reformatting the whole thing.

jandrieu:
> Why does this remove the "expandRelativeUrls" and "verificationRelationship" options? Is this intentional or an accident?

Because I got rid of the "dereferencer" as a library and the resolver should return the authoritative DID document. 

I'd be on board with defining these algorithms for a resolver client to apply after getting the DID document from resolution, but in this proposal, the software formerly known as a dereferencer is the resolving client and they don't have to return anything. They bring the resource into the current context, as is appropriate for that context. 

It's only because there's a "function" in the current spec that it needs to return things. But I think it is important to clarify what a resolving client does without requiring it to support external interfaces it has no reason for. Just like we don't require browsers to expose a dereferencing function for URLs. They just dereference into current context, as appropriate. CSS links get added to the CSS object model for the HTML page that has the link. Images are displayed. JavaScript elements are parsed and invoked. What is done with the resource is up to the client and trying to squeeze that through an interface is just confusing and weirdly limiting.

peacekeeper:
@jandrieu I still don't get why you think that the current specifcation defines a dereferencer as a library or a piece of software that needs to "expose" something.

Dereferencing is a process/function which returns a result. The word "return" is meant in an abstract/logical way, just like the function of adding 1+1 "returns" the result 2. This doesn't mean that there is an external endpoint. When a browser dereferences an HTTPS URL including a fragment, then that dereferencing process/function also "returns" something.

If you look e.g. at [Figure 10](https://www.w3.org/TR/did-resolution/#figure-client-side-dereferencing-1) of the current specification, it shows exactly what everybody is saying all the time, i.e. that dereferencing the fragment is done on the client side. I think this is exactly the same idea that you call "dereference into current context".

So I really don't understand where your confusion or desire for such major rewrites comes from?

jandrieu:
> @jandrieu I still don't get why you think that the current specifcation defines a dereferencer as a library or a piece of software that needs to "expose" something.

Because the spec requires that dereferencers expose a function with specific inputs and outputs. Section 5 DID URL Dereferencing defines a function which MUST be implemented and MUST implement a particular interface.

``` 
dereference(didUrl, dereferenceOptions) →
   « dereferencingMetadata, contentStream, contentMetadata »
```

> Dereferencing is a process/function which returns a result. 

This is the heart of the disagreement. As my suggested PR puts it, 
> This algorithm may or may not return one or more values and may affect program state, depending on the context of use. For example, dereferencing in the context of web browsing may support any number of common URL usage patterns.

Dereferencing is an algorithm, which may affect the current context in any number of ways. Requiring that it return a content stream leaves the actual integration of that string into the current context to an underspecified "client" which is rather confusingly described:
> client
> Software and/or hardware that invokes a [DID resolver](https://www.w3.org/TR/did-core/#dfn-did-resolvers) in order to execute the [DID resolution](https://www.w3.org/TR/did-core/#dfn-did-resolution) and/or [DID URL dereferencing](https://www.w3.org/TR/did-core/#dfn-did-url-dereferencing) algorithms. This invocation is done via a [binding](https://w3c.github.io/did-resolution/#dfn-binding). The term [client](https://w3c.github.io/did-resolution/#dfn-client) does not imply any specific network topology.

This language implies the "client" == the software calling the resolver (which would be the dereferencer, NOT the client calling dereferencing. So, where do we define the client that calls the dereferencer?

I believe this is the heart of the confusion about dereferencing versus resolution. It's a strange anti-conflation between clients and dereferencers, which should be the same but are confusingly given different tasks at different times. 

> Specifically, when a [DID URL](https://www.w3.org/TR/did-core/#dfn-did-urls) with a [DID fragment](https://www.w3.org/TR/did-core/#dfn-did-fragments) is dereferenced, then [Dereferencing the Resource](https://w3c.github.io/did-resolution/#dereferencing-algorithm-resource) is done by the [DID resolver](https://www.w3.org/TR/did-core/#dfn-did-resolvers), and [Dereferencing the Fragment](https://w3c.github.io/did-resolution/#dereferencing-algorithm-fragment) is done by the [client](https://w3c.github.io/did-resolution/#dfn-client).

I believe we've had some consensus that the resolver does not return the resource (except in the degenerate case where the resource is the DID document, but the caller doesn't know yet if that's the final resource without further evaluation, so its more accurate to say the resolver returns a DID document which may or may not be the final resource). So "dereferencing the resource" is definitely not what the resolver does. It returns the authoritative DID document. Getting the actual resource is the job of the client of the resolver, not the resolver.

In fact, [section 5.4.1](https://w3c.github.io/did-resolution/#dereferencing-algorithm-resource) explicitly describes what the dereferencer must do, NOT the resolver.
>  Validate that the input [DID URL](https://www.w3.org/TR/did-core/#dfn-did-urls) conforms to the did-url rule of the [DID URL Syntax](https://www.w3.org/TR/did-core/#did-url-syntax). If not, the [DID URL dereferencer](https://www.w3.org/TR/did-core/#dfn-did-url-dereferencers) MUST return the following result:

So, if your prior statement were correct, it would be in conflict with Section 5.4.1 which clearly defines [Dereferencing the Resource] as something the DID URL dereferencer does.

This is inconsistent and confusing.

> The word "return" is meant in an abstract/logical way, just like the function of adding 1+1 "returns" the result 2. This doesn't mean that there is an external endpoint. When a browser dereferences an HTTPS URL including a fragment, then that dereferencing process/function also "returns" something.

Respectfully, it is described much more explicitly as a function that MUST be implemented with a particular signature and responses. That's the problem. That definition, which is required for all conformant "dereferencers" does not fully dereference the URL into the current context, it returns a content stream that the client--who is actually dereferencing a DID URL into the current context--who must finish dereferencing. So the function as defined doesn't actually finish dereferencing, which is left to the "client" of the dereferencer. This is confusing.

> If you look e.g. at [Figure 10](https://www.w3.org/TR/did-resolution/#figure-client-side-dereferencing-1) of the current specification, it shows exactly what everybody is saying all the time, i.e. that dereferencing the fragment is done on the client side. I think this is exactly the same idea that you call "dereference into current context".

Yes, dereferencing the fragment is done by the client. So a fundamental portion of dereferencing cannot be achieved by the dereferencer. This is confusing. Why name a piece of software as a dereferencer if it doesn't actually dereference into the current context?

> So I really don't understand where your confusion or desire for such major rewrites comes from?

Hopefully my previous comments highlight specific instances of confusion. 

However, my rewrite is not from a desire to do so, but a commitment to answer Jeffrey's question from the TAG review.

First, [Jeffry Yaskin in the TAG review](https://github.com/w3ctag/design-reviews/issues/1157#issuecomment-3439062470):
> The use of "resolution" here seems inconsistent with other uses. In particular, I don't think it's "the process of determining an access mechanism and the appropriate parameters necessary to dereference a URI" from https://www.rfc-editor.org/rfc/rfc3986.html#section-1.2.2, and it's not "the process of resolving a URI reference within a context that allows relative references so that the result is a string matching the <URI> syntax rule" from https://www.rfc-editor.org/rfc/rfc3986.html#section-5 (which is also used in https://www.w3.org/TR/did-1.1/#relative-did-urls). Is it the right term? Perhaps you're just defining how to "dereference" both DIDs and DID URLs?
There's a note that 'The difference between "resolving" a DID and "dereferencing" a DID URL is being thoroughly discussed', but the link doesn't point to a comment that discusses it. This should be resolved before publication.

In fact, I do think we mean resolution in the meaning in RFC3986 as referred to by Jeffrey. But our discussion of it is confusing. So I took on responding to Jeffrey. I had hoped that would be a simple clarification of a few sentences here and there, but once I got into it, I realized that the best way to get my concerns across was to present a fully-formed alternative.

Second, Stephen Curran, on multiple occasions has argued that the distinction between dereferencing and resolution doesn't make any sense to him and that we should just use the same term for both. That one of our own struggles to distinguish between the two is a fairly clear sign that our language isn't clear. It doesn't tell us how to fix it, but it does point to a deep conceptual divide in our own community.

Third, I tried to help Stephen separate dereferencing from resolution in PR #260. First commenting piecewise on the language in those PRs. Then, at his request, I attempted a rewrite to give him some language that would address my concerns while trying to support the feature he is trying to get into the spec. 

I failed. 

I got through an update of the initial section that introduced definitions for PathService and Relative Ref, but once I started looking at the dereferencing algorithm, I too found that it was too confusing to integrate Stephen's algorithm in a way that wouldn't break existing practice.

So, after Jeffrey, Steven, and I were unable to make sense of how the current specification handles the distinction between dereferencing and resolution, I first confirmed my understanding of the algorithm, using the context of the DID Resolution Threat Model to get feedback on the process as I understand it. https://github.com/w3c/did-resolution/issues/226#issuecomment-4006405095 and https://github.com/w3c/did-resolution/issues/226#issuecomment-4006126008

Then I turned my attention back to the spec and realized that the algorithm as I understood it was, in fact, not the algorithm defined in the spec, which introduces a required component called a dereferencer, which paradoxically, can't actually dereference into the client context.

Ah-hah! I believe that this is where the confusion is. If you have to dereference the result of the dereferencer into the current context, then why is the dereferencer called a "dereferencer" and what is it really doing? I raised this as issue #310, which so far only you, Markus have responded to, with a very reasonable concern that it is late in the timeline to consider such far reaching changes.

Unfortunately, I don't know how to respond to Jeffrey's (and Stephen's and my) confusion without a revision that more clearly distinguishes the phases in the dereferencing algorithm and restores the client as the dereferencer and removes the intermediary that seems to be the root of the confusion.

So I provided spec text that I believe clears up the confusion. 

This is a big refactoring, both in substance (removing the dereferencer) and in structure (making the algorithm clearer at the top level with better separation between stages).

It is late in the game to consider it. But we didn't realize how problematic it was until Jeffry and then Stephen brought the conflict to our attention.

This PR is simply my best effort to resolve the confusion that the current spec creates.

index.html
TallTed: OK. This suggestion has correct highlighting of changes relative to your content. I have hopes that after merging it, the diff highlighting relative to the current `main` branch will be correct, and then the last line of this suggestion (which is very long, because that's what GitHub's diff tool likes) can be linted, then to also show the correct highlighting relative to the current `main`
```suggestion
Decentralized identifier (DID) resolution is the process of obtaining
the authoritative DID document and associated metadata, taking into
account any query parameters that affect DID document selection. The
resulting DID document contains information that enables
cryptographically verifiable interactions with the DID controller
related to the DID subject, including, e.g.,
cryptographic public keys. This specification covers the algorithms and guidelines to be used for DID resolution and dereferencing DID URLs, relying on the core DID specification [[[DID]]] for DID and DID URL syntax and the DID document data format.
			</p>
```

index.html
```diff
@@ -174,59 +170,136 @@
...
+	<p>
+		Resolution acquires the
+		authoritative metadata that enables resource retrieval.
+		Resolving an HTTP URL means resolving the authority part
+		(typically a host or domain name) to get the IP address to use
+		for the HTTP request. Resolving a DID URL gets the DID document 
+		and associated metadata that defines the
+		cryptographic material and service endpoints for securely
+		interacting with the associated resource.
+		The client dereferencing the DID URL uses the result returned from resolution to retrieve the actual resource and apply it to the
```
```suggestion
		The client dereferencing the DID URL uses the result returned from
		resolution to retrieve the actual resource and apply it to the
```

index.html
```diff
@@ -174,59 +170,136 @@
...
+		current computational context, either by displaying the resulting
+		resource or extracting information from the DID document to
```
```suggestion
		resource or by extracting information from the DID document to
```

index.html
```diff
@@ -174,59 +170,136 @@
...
+		interactions away, enabling any DID-enabled application to rely
+		on a consistent API, regardless of how the method interacts with
+		a VDR.
+	</p>
+	<p>
+		DIDs can be used as identifiers for both subjects and issuers of
+		Verifiable Credentials and can anchor any number of
+		cryptographic
+		verification relationships, such as assertion, authentication, capability invocation and delegation, and key agreement.
+		encryption, and key agreement.
```
```suggestion
		cryptographic verification relationships, such as assertion,
		authentication, capability invocation and delegation, and key
		agreement.
```

index.html
```diff
@@ -174,59 +170,136 @@
...
+  	<li>if the final resource is a DID Document, interpret the DID
+		document to evaluate verification methods or other properties.
+	</li>
+	</ol>
+	<p>The resolving client MAY choose any of these options
+		depending on the context in which the DID URL is used. </p>
+
+	<p>For example, the following html snippet would cause the
+		associated image to be shown in the browser displaying the
+		html page containing it.
```
```suggestion
	<p>For example, the following HTML snippet would cause the
		associated image to be shown in the browser displaying the
		HTML page containing it.
```

index.html
```diff
@@ -174,59 +170,136 @@
 
 <section id="introduction">
 <h1>Introduction</h1>
+	<p>
+		DID Resolution is the first step in dereferencing a DID URL.
+	</p>
+	<p>
+		Client software dereferences identifiers to retrieve the value of a
+		reference and bring it into the current computational context. Dereferencing a DID URL is what DID-enabled applications do to bring the referent of that DID URL into the current context.
```
```suggestion
		reference and bring it into the current computational context.
		Dereferencing a DID URL is what a DID-enabled application does to
		bring the referent of that DID URL into the current context.
```

index.html
```diff
@@ -174,59 +170,136 @@
...
+		<li>further dereference the DID URL to retrieve a secondary
+			resource&mdash;this is necessary when a DID URL contains a
+			fragment or when a web server returns a redirect;</li>
+		<li>return the DID document to another process, </li>
+	  <li>display the final dereferenced result to the current user,
+		using the resource's media type, or</li>
+  	<li>if the final resource is a DID Document, interpret the DID
+		document to evaluate verification methods or other properties.
+	</li>
+	</ol>
```
TallTed: I don't know why this is so confusing to GitHub's diff process... I've changed indent characters from tabs to spaces, revised the line lengths, and tweaked the text and punctuation.
```suggestion
    <p>With that DID document, clients can do any combination of zero or more
       of the following:</p>
    <ol>
        <li>Further dereference the DID URL to retrieve a secondary resource.
            This is necessary when a DID URL contains a fragment or when a web
            server returns a redirect.</li>
        <li>Return the DID document to another process.</li>
        <li>Display the final dereferenced result to the current user, using
            the resource's media type.</li>
        <li>If the final resource is a DID Document, interpret the DID
            document to evaluate verification methods or other properties.</li>
    </ol>
```

index.html
```diff
@@ -174,59 +170,136 @@
...
+
+	<pre class="example nohighlight"
+		title="A DID URL used to retrieve and display an image.">
+				&lt;img src="did:example:abc/image.png?service=pathService"&gt;
+				</pre>
+
+	Alternatively, a verifier working with the following [=verifiable
+	credential=] proof property would evaluate the proof using the
+	<code>#key-1</code> verification method from the DID document returned from passing the BASE DID URL <code>did:example:abc</code> to a resolver.
+	.
```
```suggestion
    <p>
        Alternatively, a verifier working with the following [=verifiable credential=]
        proof property would evaluate the proof using the <code>#key-1</code>
        verification method from the DID document returned from passing the BASE
        DID URL <code>did:example:abc</code> to a resolver.
    <p>
```

index.html
```diff
@@ -237,17 +310,14 @@ <h2>Implementer Overview</h2>
 		</p>
 
 		<p>
-			A <dfn class="lint-ignore">conforming DID URL dereferencer</dfn> is any
-			algorithm realized as software and/or hardware that complies with the relevant
+			A <dfn class="lint-ignore">conforming DID URL client</dfn> is any application or algorithm realized in software and/or hardware that complies with the relevant
 			normative statements in <a href="#dereferencing"></a>.
```
```suggestion
            A <dfn class="lint-ignore">conforming DID URL client</dfn> is any
            application or algorithm realized in software and/or hardware that
            complies with the relevant normative statements in
            <a href="#dereferencing"></a>.
```

index.html
```diff
@@ -272,9 +342,7 @@ <h2>
 		  Use Cases
 		</h2>
 		<p>
-The DID resolution specification is intended to support a broad range of use cases
-by defining a standardized interface to resolve DIDs and dereference DID URLs
-independent of the DID method of any particular DID. These usecases include:
+The DID resolution specification is intended to support a broad range of use cases including:
```
```suggestion
The DID resolution specification is intended to support a broad range of use
cases, including the following:
```

index.html
```diff
@@ -347,12 +413,17 @@ <h2>DID Parameters</h2>
...
+			method</a> and intended to function the same way for all
+		<a>DIDs</a>. Other DID parameters
+		are not supported by all <a>DID methods</a>. Where optional
+		parameters are
+		supported, they are expected to operate uniformly across the
+		<a>DID methods</a>
+		that do support them. The following table provides common DID
+		parameters that SHOULD function the same way across all <a>DID
+			methods</a>. Support for all
 		<a href="#did-parameters">DID Parameters</a> is OPTIONAL.
```
TallTed, line 435:
```suggestion
            Some DID parameters are completely independent of any specific
            <a>DID method</a> and intended to function the same way for all
            <a>DIDs</a>. Other DID parameters are not supported by all
            <a>DID methods</a>. Where optional parameters are supported,
            they are expected to operate uniformly across the <a>DID
            methods</a> that do support them. The following table provides
            common DID parameters that SHOULD function the same way across
            all <a>DID methods</a>. Support for each and every
            <a href="#did-parameters">DID parameter</a> is OPTIONAL.
```

index.html
```diff
@@ -997,94 +1012,103 @@ <h2>DID Resolution Algorithm</h2>
 </section>
 
 <section id="dereferencing">
-	<h1>DID URL Dereferencing</h1>
+	<h1><dfn data-lt="DID URL Client|Client">DID URL Client</dfn></h1>
```
swcurran, line 1023: I think this "DID URL Client" algorithm is significantly under-specified such that an implementers will not know what to do. The following are a few examples that I think are problematic:

- Combinations of service and serviceType parameters. For example, as specified serviceType parameters are silently ignored if a  service parameter exists.
- I don't see what happens when a service or serviceType is "selected"?  It’s not clear to me what is the result of that selection.
- Per the algorithm, if a property in the DIDDoc or DIDDoc Metadata is found that relates to dereferencing it should be used  — without any trigger to indicate it should be used. I would interpret that as, if I pass in a DID URL that is “just a DID”, and I find a Linked Resource in the retrieved DID Doc, I am supposed to return the resource it points to.  I’m sure that is not what is meant, but that is how I read it.
- I don’t see that anything is defined if any other query parameters are encountered, or if a path is included.

While these are examples, I don't think this algorithm is backwards compatible with the existing spec (is that a goal?) and is incomplete.

In general, I think this algorithm is a step back from what is in the spec today, making implementations harder.

swcurran:
I’m not a fan of how the term “client” is used in this PR. In my mind, the goal of this specification is that a DID URL can be passed to different implementations of this specification, and the result returned will be the same from each.  I think of the “client” is the thing that calls those implementations, and what gets back what it expects based on the DID URL it passed in and then using it in some way — via business logic that is outside the scope the specification. While I have struggled with the term “DID URL Dereferencer”, I think it separates it from client and should be retained. I would prefer not use the term “client” as it is used in this PR.

index.html
wip-abramson: `did:btcr2:k1abcd?serviceType=SMTBeacon`

Here, the resource I care about is the JSON that defines all services in the DID document of a specific type.

Is this not a valid use? It was possible in the previous definition.

index.html
```diff
@@ -1099,517 +1123,327 @@ <h1>DID URL Dereferencing</h1>
...
+					`did:ex:abc#key-1` becomes `did:ex:abc` .</li>
+				<li><strong>Prepare</strong> <var><a>resolution
+							options</a></var>. See the [[resolve]] algorithm for
+					options.</li>
+			</ol>
+		</section>
+		<section id="dereferencing-algorithm-resolve-the-did-url">
+			<h3>Resolve the DID URL</h3>
+			<ol class="algorithm">
+				<li><strong>Call</strong> the [=resolve=] function on the
```
wip-abramson, line 1209: There still feels like a number of ReSpec inconsistencies throughout this PR. These would be good to fix.

See these linking shorthand table in the ReSpec doc - https://respec.org/docs/#linking-shorthands

Are you intending to reference a term `resolve` in this line? 

I think you are intending to link to the resolve function defined in the DID Resolution section. 

If so, this might be one way to achieve that.

```suggestion
				<li><strong>Call</strong> the <a href="#resolving">resolve</a> function on the
```

index.html
```diff
@@ -1099,517 +1123,327 @@ <h1>DID URL Dereferencing</h1>
...
+					a [[local resolver]], which does not (likely a local
+					resolver will be
+					invoked using standard linked library or interprocess
+					communication). Implementers MUST enable users to choose the
+					resolvers they trust for specific DID methods.</li>
+				<li><strong>Remove</strong> any fragment part of the DID
+					URL to create a [=base DID URL=], e.g.,
+					`did:ex:abc#key-1` becomes `did:ex:abc` .</li>
+				<li><strong>Prepare</strong> <var><a>resolution
+							options</a></var>. See the [[resolve]] algorithm for
```
wip-abramson: `[[SPEC]]` is an informative biblio reference - https://respec.org/docs/#linking-shorthands.

Again, I think what is meant here is a relative link to the DID Resolution section of the spec.

```suggestion
							options</a></var>. See the <a href="#resolving">resolve</a> algorithm for
```

There are a number of instances throughout this PR that appear to use incorrect types of linking. Could you review? 

All red squiggly lines in the preview are a good place to start: https://pr-preview.s3.amazonaws.com/LegReq/did-resolution/pull/315.html

index.html
```diff
@@ -1099,517 +1123,327 @@ <h1>DID URL Dereferencing</h1>
...
+				<li><strong>Select</strong> a resolver that is capable of
+					handling the DID's method. This could be a [[remote
+					resolver]], which would use HTTPS for communications, or
+					a [[local resolver]], which does not (likely a local
+					resolver will be
+					invoked using standard linked library or interprocess
+					communication). Implementers MUST enable users to choose the
+					resolvers they trust for specific DID methods.</li>
+				<li><strong>Remove</strong> any fragment part of the DID
+					URL to create a [=base DID URL=], e.g.,
```
wip-abramson, line 1199: I believe this is correct usage - you are referencing the term `base DID URL`. However, this term does not yet exist. I know you have mentioned this before, but perhaps you can remind me.

Why not just define this term in the terms.html as part of this PR?

I think you mentioned xref, this is how we can cross link to terms defined in other specifications - https://respec.org/docs/#xref. Are there terms in this PR that should be defined in other specs?

wip-abramson (changes requested):
Currently there are a number of ReSpec linking inconsistencies that should be addressed. 

At least the instances where the type of ReSpec linking used appears incorrect. E.g. a biblio link instead of a section

index.html
jandrieu: That's a fine use. Why wouldn't that be allowed in this PR?

The resolving client gets back a DID document. Their analysis of the document may suggest that there are multiple resources to choose from. And they do. Choose.

The end result of dereferencing is the selection of a single resource. Here's the pull quote from the MDN reference Manu shared in the meeting (https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/300):

> The HTTP 300 Multiple Choices [redirection response](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status#redirection_messages) status code indicates that the request has more than one possible response. **The user-agent or the user should choose one of them.**

> Note: In [agent-driven content negotiation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Content_negotiation#agent-driven_negotiation), a client and server collaboratively decide the best variant of a given resource when the server has multiple variants. Most clients lack a method for automatically choosing from responses, and the additional round-trips slow down client-server interaction. [Server-driven content negotiation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Content_negotiation#server-driven_content_negotiation) is far more common, where a server chooses the most appropriate resource for the client based on the request headers ([Accept-Language](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept-Language), [Accept](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept), etc.).

> The server should include content in the response that contains a list of resource metadata and URIs from which the **user or user agent can choose**.

Emphasis is mine.

Two points. 

First of all, this is a REDIRECT code. Not the returning of a resource.

Second, the client, necessarily CHOOSES one of the possible redirects. One of them. Which it then dereferences.

At the end of the process, only a single referent is dereferenced into the current context.

This, again, illustrates the unnecessary complexity of a "dereferencer" which is not actually able to dereference.

What is being asked for here is a form of redirect, not dereferencing.

jandrieu:
> If I'm reading this correctly, it seems "relativeRef" is only mentioned in the "Service Selection Algorithm", not in the "Service Type Selection Algorithm"?
> 
> Also, I wonder about this sentence:
> 
> > If the input [DID URL](https://www.w3.org/TR/did-core/#dfn-did-urls) contains the [DID parameter](https://pr-preview.s3.amazonaws.com/LegReq/did-resolution/pull/315.html#did-parameters) relativeRef and a service of type "PathService" is selected ...
> 
> Does this mean that "relativeRef" is only supported in conjunction with PathService, not with "service" and "serviceType" parameters?

Not exactly. If another service type wants to use relative ref, that's fine too. I was trying to split the difference with @swcurran's PR and support the only service type I know of that uses relative ref. It certainly is the only one proposed so far to be defined in this specification. 

What should happen if relative ref is defined and there is no service endpoint known to use that query parameter? Happy to update if we figure out how we want to handle that.

Another thing came up when @wip-abramson  and I discussed this.

One consequence of this update to the algorithm is that it becomes exceptionally clear that when dereferencing a DID URL with a service or serviceType query parameter, that the service definition needs to define a retrieval strategy.

And none of the service types I know adequately address this. There was a long held assumption by many of us--and I also bought into it--that a service is likely derereferenced by applying a GET request of the http URL in the service_endpoint property. But that sort of tribal assumption is one of the things we need to formalize when we define a spec like this one.

That approach doesn't work if the service endpoint isn't an HTTPS URL. You can see this is the challenge with the bitcoin: scheme used in the endpoints for BTCR2. A "standard" dereferencing of that endpoint is just going to hurl. If fact, according to BIP 21 https://en.bitcoin.it/wiki/BIP_0021 that URI is used for making bitcoin payments and when dereferenced in a wallet brings up a payment form prefilled with that bitcoin address. This is clearly not how you retrieve the data from a Beacon as defined in BTCR.

The BTCR2 specification doesn't talk about how to dereference a DID URL with a service query for a beacon (from example at https://dcdpr.github.io/did-btcr2/data-structures.html#initial-did-document-example-panel-show) :
```did:btcr2:k1q5pvh5zask8khdg7p58ygveewkcufetu3dlqyaca5dzqct6mjhf540qhrxgv3?service=%23initialP2PKH```

In theory that SHOULD dereference to something, but what is undefined.

Instead, that URL is actually used as a URI by the resolution algorithm and is really, only ever used when validating the chain of updates as described in resolution.

Its "retrieval strategy" when accessed directly is undefined.

Which is, IMO, the right boundary. If a dereferencing client gets a URL with that service, but it doesn't know how to retrieve the resource for a "SingletonBeacon" it should error out.

index.html
jandrieu: Happy to update the references if we go this direction. I think this the most minor of nits. Easy to fix if we want to.

index.html
jandrieu: As I mentioned in our call, Will. I'm happy to figure out how to make the references work whether local or through xref should we want to go this direction. 

This PR is an attempt to provide an alternative to Stephen's because none of my direct comments on that were getting traction. Fixing references is an easy fix.

jandrieu:
> I’m not a fan of how the term “client” is used in this PR. In my mind, the goal of this specification is that a DID URL can be passed to different implementations of this specification, and the result returned will be the same from each. I think of the “client” is the thing that calls those implementations, and what gets back what it expects based on the DID URL it passed in and then using it in some way — via business logic that is outside the scope the specification. While I have struggled with the term “DID URL Dereferencer”, I think it separates it from client and should be retained. I would prefer not use the term “client” as it is used in this PR.

Yes. We share that goal. 

But the term "client" is not one I created. It was already in the specification, underdefined as just the "client of the resolver". 

A URL (did or otherwise) can be passed to any application that wants to consume it, but what actually happens with it is entirely context-dependent. 

A URL in an RDF statement is used differently than a URL in a hyperlink is different than a URL in a Verifiable Credential.

Even within the browsing context, what is expected to be consistent is that all browsers will attempt to render the same resource (or really a representation of a facet of that resource) by querying the same authority and treating its response as authentic. But different browsers handle different media types differently. Some browsers--like Brave--intentionally do NOT dereference many of the tracking mechanisms that most browsers blindly process, which absolutely breaks many web pages, even when Brave *is* displaying exactly the same HTML as another browser. The commitment is not the equivalence of the rendering, but the equivalence of the network request to retrieve a representation of the resource from a server specified by the authority (which includes CDNs and round-robin DNS approaches).

> I think of the “client” is the thing that calls those implementations, and what gets back what it expects based on the DID URL it passed in and then using it in some way — via business logic that is outside the scope the specification. 

Yes. That's exactly what the client is. The client, which is the software that calls the resolver, is the one that processes the business logic to apply the result of resolution to the current context. There is no dereferencer needed in that flow and plenty of reasons it creates problems, which I have enumerated.

index.html
jandrieu: > I think this "DID URL Client" algorithm is significantly under-specified such that an implementers will not know what to do. 

That may still be true. However, I contend that it is far clearer what to do in this PR than in the current specification. 

> The following are a few examples that I think are problematic:
> 
> * Combinations of service and serviceType parameters. For example, as specified serviceType parameters are silently ignored if a  service parameter exists.

That's an interesting question. My personal opinion is that specifying both is overspecifying and I would return an error. @wip-abramson suggested maybe the better interpretation is that if both are used that a consistency check is applied such that if the service type doesn't match the service object specified by the service parameter, return an error.

What we should NOT do is simply return all the URLs for all the matching objects. How would the client understand which one was for which? 

My algorithm has a specific prioritization. It could probably be improved, but IMO, if a service is named in a service parameter, that should define retrieval. If that service name fails to find any objects, the dereferencing should fail. A clear deterministic algorithm will simplify implementations. An algorithm that tries to waffle on how to prioritize will not.

> * I don't see what happens when a service or serviceType is "selected"?  It’s not clear to me what is the result of that selection.

From the PR:
> The retrieval strategy is determined by the type property of the selected service. 

This is explicitly stated in the Section 6.3 Determine Retrieval Strategy steps 1 and 2.

> * Per the algorithm, if a property in the DIDDoc or DIDDoc Metadata is found that relates to dereferencing it should be used  — without any trigger to indicate it should be used. I would interpret that as, if I pass in a DID URL that is “just a DID”, and I find a Linked Resource in the retrieved DID Doc, I am supposed to return the resource it points to.  I’m sure that is not what is meant, but that is how I read it.

That is what is meant. We have an unresolved question about how the client knows which properties have that feature. My proposal is not in this PR, but I have proposed it in yours, Stephen. Define a type property that labels an object as a path handler.

If there's support for that, I could add it.

In fact, I have said many times that supporting the actual functionality you want @swcurran is going to be easier in this PR than in yours, which breaks linkedResources in ways that I still haven't succeed at explaining to you. I created that property. I know why it was designed the way it was, and there simply is no way to return a URL, did or otherwise, for all of the variants that linkedResources supports.

> * I don’t see that anything is defined if any other query parameters are encountered, or if a path is included.

Indeed. I don't know of any impact on dereferencing from any other query parameters. The ones defined in the spec like `versionId` and `versionTime` do not affect dereferencing, only resolution. Personally, I would rather NOT open up the dereferencing algorithm to dynamically change based on arbitrary query parameters other than `service` and `serviceType`. We have enough of a challenge handling the arbitrary properties that have their own dereferencing algorithms.

So, I didn't include it as I know of no approach to affecting dereferencing through query parameters other than service and serviceType. And I strongly think it would be a bad idea to extend beyond that. If we have evidence of practices in the wild that do in fact, do that, then I might be convinced, but please, let's not add complexity for some notion of consistency.
 
> While these are examples, I don't think this algorithm is backwards compatible with the existing spec (is that a goal?) and is incomplete.

I'm sure it can be made more complete. I don't pretend that it is done. This is literally the first pass at a MAJOR change and if the group doesn't want to go in this direction, there's no need to improve it.

So, no, it was not designed to be compatible with the existing spec. It was designed to clear up the confusion that the existing spec has created. For you. For Jeffrey. For me. And, I believe, for a lot of folks. 

Forcing dereferencing into a function call is a category error that, IMO, is adding exceptional complexity and confusion without actually creating any clear features or functions we don't already get by defining a coherent algorithm the client uses.

> In general, I think this algorithm is a step back from what is in the spec today, making implementations harder.

It may make them harder, but I think it makes them clearer. What is hard is figuring out what to do with the five different kinds of results you might get from the dereferencer and even harder to figure out how to dereference a URL whose method you don't understand. And that latter part is, IMO, the entire point of the DID resolution spec: to allow arbitrary clients to get a DID document they can reason over within their particular context, regardless of the DID method. Once the client software needs to do things differently because of the DID method, we fail at interop.

I would welcome a PR from anyone else if folks see an easier way to respond to Jeffrey's confusion. At this point, it seems that I'm the only one who has been willing to take that on and it is more than frustrating to have my work attacked because it changes things. The POINT is to clear up the confusion. That means changes. Based on my read of the spec, that dereferencer is the cause and change is called for.

If you think otherwise, anyone, please suggest an alternative PR that you feel will clear up Jeffrey's concerns.

peacekeeper (changes requested):
In general I like the idea of making the dereferencing process more modular, but for now:

1. As I mentioned in https://github.com/w3c/did-resolution/pull/315#issuecomment-4180026704, I think this PR breaks how service, serviceType, and relativeRef work at the moment.

2. Also, several of my comments in https://github.com/w3c/did-resolution/pull/311 have not been addressed and still apply here in this PR.

3. I disagree with removing several features such as verificationRelationship and expandRelativeUrls, which we have extensively discussed in the past and decided to include.

4. This removes the dereference() function, and I worry that this might break the use cases I listed in https://github.com/w3c/did-resolution/issues/306#issuecomment-4076381255.

jandrieu:
> In general I like the idea of making the dereferencing process more modular, but for now:
> 
> 1. As I mentioned in [#315 (comment)](https://github.com/w3c/did-resolution/pull/315#issuecomment-4180026704), I think this PR breaks how service, serviceType, and relativeRef work at the moment.
> 2. Also, several of my comments in #311 have not been addressed and still apply here in this PR.
> 3. I disagree with removing several features such as verificationRelationship and expandRelativeUrls, which we have extensively discussed in the past and decided to include.
> 4. This removes the dereference() function, and I worry that this might break the use cases I listed in [#306 (comment)](https://github.com/w3c/did-resolution/issues/306#issuecomment-4076381255).

I'll take a look at these in more detail and attempt to address what I can.

PR #318, opened 2026-04-09
Clarify that "DID resolution" is consistent with "URI resolution"

peacekeeper:
This addresses https://github.com/w3c/did-resolution/issues/226, by clarifying that the term "DID resolution" is consistent with "URI resolution" as defined in RFC3986, and that it is a part of the "DID URL dereferencing" process.

***
<a href="https://pr-preview.s3.amazonaws.com/peacekeeper/did-resolution/pull/318.html" title="Last updated on Apr 17, 2026, 8:42 PM UTC (0463af0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-resolution/318/ea94ecc...peacekeeper:0463af0.html" title="Last updated on Apr 17, 2026, 8:42 PM UTC (0463af0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -478,10 +478,10 @@ did:example:123?service=files&amp;relativeRef=/resume.pdf
 		<a href="#dereferencing"></a>.
 	</p>
 	<p class="note">
-		In order to be able to dereference a DID URL, one first needs a <a>DID document</a>,
+		To be able to dereference a DID URL, one first needs a <a>DID document</a>,
 		which is the result of the <a>DID resolution</a> process. The <a>DID document</a>
 		is used in various ways during the <a>DID URL dereferencing</a> process.
-		The term <a>DID resolution</a> is therefore consistent with "URI resolution" as defined in
+		The term <a>DID resolution</a> is consistent with "URI resolution" as defined in
 		<a data-cite="RFC3986#section-1.2.2">RFC3986 Section 1.2.2</a>, since it provides
 		"the appropriate parameters necessary to dereference a URI".
 	</p>
--- a/index.html
+++ b/index.html
@@ -473,6 +473,16 @@ did:example:123?service=files&amp;relativeRef=/resume.pdf
 		document</a> by using the "Resolve" operation of the applicable <a>DID method</a>
 		as described in <a data-cite="did-core#method-operations">Method Operations</a>.
 	</p>
+	<p>
+		<a>DID resolution</a> is a part of <a>DID URL dereferencing</a>, as described in
+		<a href="#dereferencing"></a>.
+	</p>
+	<p class="note">
+		The term <a>DID resolution</a> is consistent with "URI resolution" as defined in
+		<a data-cite="RFC3986#section-1.2.2">RFC3986 Section 1.2.2</a>, i.e.
+		"the process of determining an access mechanism and the appropriate parameters
+		necessary to dereference a URI".
+	</p>
 	<p>	All [=conforming DID resolvers=] implement the function below, which has the
 		following abstract form:
 	</p>
--- a/index.html
+++ b/index.html
@@ -478,10 +478,12 @@ did:example:123?service=files&amp;relativeRef=/resume.pdf
 		<a href="#dereferencing"></a>.
 	</p>
 	<p class="note">
-		The term <a>DID resolution</a> is consistent with "URI resolution" as defined in
-		<a data-cite="RFC3986#section-1.2.2">RFC3986 Section 1.2.2</a>, i.e.
-		"the process of determining an access mechanism and the appropriate parameters
-		necessary to dereference a URI".
+		In order to be able to dereference a DID URL, one first needs a <a>DID document</a>,
+		which is the result of the <a>DID resolution</a> process. The <a>DID document</a>
+		is used in various ways during the <a>DID URL dereferencing</a> process.
+		The term <a>DID resolution</a> is therefore consistent with "URI resolution" as defined in
+		<a data-cite="RFC3986#section-1.2.2">RFC3986 Section 1.2.2</a>, since it provides
+		"the appropriate parameters necessary to dereference a URI".
 	</p>
 	<p>	All [=conforming DID resolvers=] implement the function below, which has the
 		following abstract form:


wip-abramson:
So reviewing #226 and the original comment raised by @jyasskin here https://github.com/w3ctag/design-reviews/issues/1157.

I noticed that the comment was specifically about the introduction. 

Perhaps @jyasskin could comment if this addresses his concern.

index.html
```diff
@@ -473,6 +473,16 @@ <h1>DID Resolution</h1>
...
 	</p>
+	<p>
+		<a>DID resolution</a> is a part of <a>DID URL dereferencing</a>, as described in
+		<a href="#dereferencing"></a>.
+	</p>
+	<p class="note">
+		The term <a>DID resolution</a> is consistent with "URI resolution" as defined in
+		<a data-cite="RFC3986#section-1.2.2">RFC3986 Section 1.2.2</a>, i.e.
+		"the process of determining an access mechanism and the appropriate parameters
+		necessary to dereference a URI".
```
jyasskin: Thanks for the ping, @wip-abramson. I wasn't around for the theoretical debates that led to the definition in RFC 3986, so I could be misunderstanding what an "access mechanism and the appropriate parameters" are meant to be. (Reading the RFC, my impression is that the authors didn't have a clear understanding themselves.) With my current understanding, I see resolution defined here as

> resolve(did, resolutionOptions) →
>   « didResolutionMetadata, didDocument, didDocumentMetadata »

If this is determining the "access mechanism and appropriate parameters", I would expect to find  those in the results. [didResolutionMetadata](https://pr-preview.s3.amazonaws.com/peacekeeper/did-resolution/pull/318.html#did-resolution-metadata) has error information, a content type, and a proof, which doesn't seem like an access mechanism or parameters. The didDocument is one of the [possible result types of deferencing](https://pr-preview.s3.amazonaws.com/peacekeeper/did-resolution/pull/318.html#dfn-contentstream), so it seems like more than just the access mechanism or parameters. And the didDocumentMetadata includes fields that describe the document, but things like 'created' don't seem necessary to dereference a URI.

Perhaps more generally, the only output of the DID Resolution algorithm that I see being used as part of DID Dereferencing is the document itself, which for simple DID URLs is passed directly through to the output.

If you do think you're matching an interpretation of the RFC3986 definition that I haven't considered, it would be nice for this note to explain that interpretation and how this use matches it, rather than just asserting that they're consistent.

----

I haven't checked this with the rest of the TAG, but one possible direction might be to declare that the RFC 3986 definition of resolution simply isn't useful: that determining an access mechanism is hopelessly intertwined with dereferencing the full URL, for basically all schemes. Then, if you feel it's important to have a distinct name for the process of finding the relevant DID Document for a given DID URL, you could use "resolution" as your own term. You could even mark it as a ["willful violation"](https://infra.spec.whatwg.org/#willful-violation) of RFC 3986, to be clear that nobody should bother making my comment in the future.

I _don't_ feel like it's important to have this distinct name. I think it would be clearer to name an algorithm "retrieve the DID Document for a DID URL", and that it's not likely to come up enough in other specs to need a short term for it. But if the WG considers the question and thinks I'm wrong about this, I'm comfortable yielding to your expertise.

dlongley: One possible way to apply RFC 3986 here:

In order to be able to reference a DID URL, one first needs a DID Document. Therefore, the potential "access mechanisms and appropriate parameters" for a DID URL is at least a DID Document.

The DID document (and in some sense, its content type, which tells you that it's a DID document) provides the parameters necessary to dereference the path of the DID URL (and any hash fragment, etc.) to its referent.

There may also be additional parameters for dereferencing that are included in the metadata about the DID document / DID URL. This might include specific time information the dereferencer will care about and even perhaps the parsed DID method -- which would allow late method-specific behaviors to be performed, but I agree with other's statements that much or all of this is generally an anti-pattern that leads to interop failure. We should try to push most or all of this into the resolution layer, not the dereferencing layer; the application / client (that will do the dereferencing) should not be concerned with it.

swcurran:
The tricky part of this:

> In order to be able to reference a DID URL, one first needs a DID Document. Therefore, the potential "access mechanisms and appropriate parameters" for a DID URL is at least a DID Document.

is that to get the intended DID Document, you have to process the DID URL, because parts of the DID URL impact that retrieval. We have to agree on how to balance that.

> We should try to push most or all of this into the resolution layer

I totally agree -- the ideal is that all post DIDDoc retrieval processing of a DID URL is non-DID method specific. Where I think care is needed is where the DID URL is the "only" identifier for a resource - such as something stored on a blockchain. We then have to use "blockchain-specific" code for retrieval -- e.g. DID Method handling. We need to have a path for that.

Your use of "layer" is, I think the core of the issue we're struggling with in the Working Group. Are there separate resolution and dereferencing "layers"?  My leaning (not hard stance) is that there are steps (not layers) in the dereferencing of a DID URL to get a consistent result from all implementations. The input is a DID URL, and there is a deterministic output from executing a process.

index.html
peacekeeper: > The didDocument is one of the [possible result types of deferencing](https://pr-preview.s3.amazonaws.com/peacekeeper/did-resolution/pull/318.html#dfn-contentstream), so it seems like more than just the access mechanism or parameters.

Yes, but the didDocument is the ONLY possible result type of **resolving**. It contains information which is used during the dereferencing process. I agree with @dlongley that "In order to be able to de(?)reference a DID URL, one first needs a DID Document. "

If we add 1-2 sentences to explain this in this PR, maybe that would solve the issue?

peacekeeper:
> the ideal is that all post DIDDoc retrieval processing of a DID URL is non-DID method specific.

Just a note that there is no consensus on this. Also, I think this is a separate topic that is unrelated to the PR here.

index.html
msporny: @jyasskin wrote:
> I don't feel like it's important to have this distinct name. I think it would be clearer to name an algorithm "retrieve the DID Document for a DID URL", and that it's not likely to come up enough in other specs to need a short term for it.

FWIW, strongly agree with the statement above and @dlongley's take as well.

We can't just assert that DID resolution is aligned, because as @jyasskin said, RFC 3986 is fairly vague in its "access mechanism and the appropriate parameters" language.

To be clear, I'm fine w/ this PR, but it doesn't feel like it clears up the core issue that @jyasskin noted in the TAG review.

msporny (approved):
Ok with the PR, in general, but don't think it addresses @jyasskin's comment.

peacekeeper:
Based on above feedback, I tried to update the language a bit to explain WHY it is consistent, rather than just asserting that it is.

Maybe @jyasskin @dlongley @msporny could re-review, to check if this addresses the issue mentioned in https://github.com/w3ctag/design-reviews/issues/1157#issuecomment-3439062470?

msporny (approved):
The revision on the original PR LGTM. Curious to see if it addresses @jyasskin and @jandrieu's concerns.

jyasskin (approved):
That fits it into the definition; thanks for working through it with me.

dmitrizagidulin (approved):
👍

dmitrizagidulin:
Multiple approvals, no objections. Should be ready to merge after the appropriate waiting period.

dlongley (approved):

ottomorac (approved):

index.html
```diff
@@ -473,6 +473,18 @@ <h1>DID Resolution</h1>
 		document</a> by using the "Resolve" operation of the applicable <a>DID method</a>
 		as described in <a data-cite="did-core#method-operations">Method Operations</a>.
 	</p>
+	<p>
+		<a>DID resolution</a> is a part of <a>DID URL dereferencing</a>, as described in
+		<a href="#dereferencing"></a>.
+	</p>
+	<p class="note">
+		In order to be able to dereference a DID URL, one first needs a <a>DID document</a>,
```
```suggestion
		To be able to dereference a DID URL, one first needs a <a>DID document</a>,
```

index.html
```diff
@@ -473,6 +473,18 @@ <h1>DID Resolution</h1>
...
+		which is the result of the <a>DID resolution</a> process. The <a>DID document</a>
+		is used in various ways during the <a>DID URL dereferencing</a> process.
+		The term <a>DID resolution</a> is therefore consistent with "URI resolution" as defined in
```
```suggestion
		The term <a>DID resolution</a> is consistent with "URI resolution" as defined in
```

PR #319, opened 2026-04-09
Explain that DID URL dereferencer acts as client of a DID resolver

peacekeeper:
This tries to address some confusion about clients, resolvers, and dereferencers, by stating explicitly that a DID URL dereferencer acts as a client of a DID resolver.

This does NOT mean that a DID resolver is a remote service; it can also be an entirely local process.

This does NOT mean that a DID URL deferencer exposes an endpoint; it can also be an entirely local process.

***
<a href="https://pr-preview.s3.amazonaws.com/peacekeeper/did-resolution/pull/319.html" title="Last updated on Apr 9, 2026, 1:58 PM UTC (52f3f3c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-resolution/319/ea94ecc...peacekeeper:52f3f3c.html" title="Last updated on Apr 9, 2026, 1:58 PM UTC (52f3f3c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1003,8 +1003,15 @@ string">ASCII string</a>.
 		The <a>DID URL dereferencing</a> function dereferences a <a>DID URL</a> into a
 		<a>resource</a> with contents depending on the <a>DID URL</a>'s components,
 		including the <a>DID method</a>, method-specific identifier, path, query, and
-		fragment. This process depends on <a>DID resolution</a> of the <a>DID</a>
-		contained in the <a>DID URL</a>. <a>DID URL dereferencing</a> might involve
+		fragment.
+	</p>
+	<p>
+		This process depends on <a>DID resolution</a> of the <a>DID</a>
+		contained in the <a>DID URL</a>. In other words, the <a>DID URL dereferencer</a>
+		acts as a <a>client</a> of a <a>DID resolver</a> during this process.
+	</p>
+	<p>
+		<a>DID URL dereferencing</a> might involve
 		multiple steps (e.g., when the DID URL being dereferenced includes a fragment),
 		and the function is defined to return the final resource after all steps are
 		completed. The following figure depicts the relationship described


index.html
```diff
@@ -1003,8 +1003,15 @@ <h1>DID URL Dereferencing</h1>
...
 		<a>resource</a> with contents depending on the <a>DID URL</a>'s components,
 		including the <a>DID method</a>, method-specific identifier, path, query, and
-		fragment. This process depends on <a>DID resolution</a> of the <a>DID</a>
-		contained in the <a>DID URL</a>. <a>DID URL dereferencing</a> might involve
+		fragment.
+	</p>
+	<p>
+		This process depends on <a>DID resolution</a> of the <a>DID</a>
+		contained in the <a>DID URL</a>. In other words, the <a>DID URL dereferencer</a>
+		acts as a <a>client</a> of a <a>DID resolver</a> during this process.
```
msporny, line 1011: Using the word "client" here is going to confuse people. Much of the "client-server" language used in IETF specs, and a number of W3C ones, really does mean "A remote system". People won't think "caller of a software library function" when they read "client".

msporny, line 1011:
```suggestion
		This process depends on <a>DID resolution</a> of the <a>DID</a>
		contained in the <a>DID URL</a>. In other words, the <a>DID URL dereferencing</a>
		process happens after, and is dependent upon, the <a>DID resolution</a> process.
```

msporny (changes requested):

index.html
peacekeeper: I agree, but in the context of DIDs, we have always been using the term "client" in this way. If you execute a local library to resolve a `did:key`, then according to the current spec, you also have a "client" which invokes a "resolver", even if it's done via a local binding. This is why in the [terminology about client](https://www.w3.org/TR/did-resolution/#dfn-client), we say this:

> The term client does not imply any specific network topology.

If we want to avoid this confusion, maybe we should use another term instead? E.g. "caller"?

peacekeeper: I'm fine with this change, but I think @jandrieu wanted to explicitly mention that a dereferencer is a "client" to a resolver, e.g. see this comment https://github.com/w3c/did-resolution/pull/311#issue-4137259083 .

swcurran: If we have used the term "client" in this way, we should stop.  Perhaps `In other words, the <a>DID URL dereferencer</a> invokes a <a>DID resolver</a> during this process.`

"Client" should be reserved for the logical external component that passes in a DID URL, gets back some sort of resource, and who handles the fragment given the resource returned.

swcurran: The use of "after" is a bit tricky here. It is **part** of the DID URL Dereferencing process if for now other reason than there may be DID parameters used in DID resolution.  It is always a DID URL passed in by the client and dereferencing must begin by resolving the DID using the DID and any relevant parameters.

PR #324, opened 2026-04-12
shifts from resolving "DIDs" to "DID URLs" 

jandrieu:
since query parameters MUST propagate and clients may make errors.

This address two of the line items in #311

2. Pass a full DID URL to resolution, since the resolver may need DID URL parameters that the dereferencer is unaware of. This is especially true for method-specific extensions where resolvers know the special magic, e.g., BTCR, but the resolving client just wants to get back the right DID document, without needing to parse the incoming URL.

7. Updated the DID resolution image to clarify that resolution always returns a DID document and that dereferencing always presents a resource (which might mean display or serialized output), noting that the resource returned from dereferencing can be the DID document

Handling 2 without 7 would have left the document inconsistent.

***
<a href="https://pr-preview.s3.amazonaws.com/LegReq/did-resolution/pull/324.html" title="Last updated on Apr 12, 2026, 7:05 PM UTC (c2897fc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-resolution/324/ea94ecc...LegReq:c2897fc.html" title="Last updated on Apr 12, 2026, 7:05 PM UTC (c2897fc)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -19,7 +19,7 @@
         shortName: "did-resolution",
 
         // subtitle
-        subtitle: "Algorithms and guidelines for resolving DIDs and dereferencing DID URLs",
+        subtitle: "Algorithms and guidelines for resolving and dereferencing DID URLs",
 
         // if you wish the publication date to be other than today, set this
         // publishDate:  "2009-08-06",
@@ -132,14 +132,13 @@
     <section id='abstract'>
       <p>
 Decentralized identifier (DID) resolution is the process of obtaining
-a DID document and accompanying metadata for a specific DID. The process
-takes a DID and a set of resolution options as its input and returns a
+a DID document and accompanying metadata for a specific DID URL. The process
+takes a DID URL and a set of resolution options as its input and returns the authoritative 
 DID document and associated metadata about the resolved document and the
 resolution request. A resolved DID document is a set of information which
 enables cryptographically verifiable interactions with the DID subject,
 including mechanisms such as cryptographic public keys. This specification
-covers the algorithms and guidelines to be used for DID resolution and
-relies on the core DID specification,
+covers the algorithms and guidelines to be used for DID resolution and dereferencing. It relies on the core DID specification,
 [[[DID]]],
 which describes the underlying DID architecture in full detail.
       </p>
@@ -175,17 +174,12 @@ which describes the underlying DID architecture in full detail.
 <section id="introduction">
 <h1>Introduction</h1>
 
-	<p>[=DID resolution=] is the process of obtaining a <a>DID document</a> for a given <a>DID</a>. This is one
-	of four required operations that can be performed on any DID ("Read"; the other ones being "Create", "Update",
-	and "Deactivate"). The details of these operations differ depending on the <a>DID method</a>.
-	Building on top of [=DID resolution=], <a>DID URL dereferencing</a> is the process of retrieving a representation
-	of a resource for a given <a>DID URL</a>. Software and/or hardware that is able to execute these processes is called
-	a <a>DID resolver</a>.
+	<p>[=DID resolution=] is the process of obtaining a <a>DID document</a> for a given <a>DID URL</a> (all <a>DIDs</a> are DID URLs). 
+	Relying on [=DID resolution=], <a>DID URL dereferencing</a> is the process of applying a representation
+	of a resource for a given <a>DID URL</a> to the current computational context. Software and/or hardware that is able to execute [=DID resolution=] is called	a <a>DID Resolver</a>. Software and/or hardware that is able to call [=Resolve=] on a DID resolver as part of dereferencing a DID URL are called DID URL clients.
 
-	<p>This specification defines a standard interface that clients can use to execute [=DID resolution=] and <a>DID URL dereferencing</a>
-		requests, independent of any specific <a>DID method's</a> "Resolve" operation that a [=DID resolver=] supports. Additionally,
-		this specification defines requirements, algorithms including their inputs and results, architectural options, and various
-		security and privacy considerations relevant to implementing a [=DID resolver=] or <a>DID URL dereferencer</a>.</p>
+	<p>This specification defines a standard interface and algorithm for [=DID Resolution=] and the algorithm that clients use for <a>DID URL dereferencing</a>, independent of any specific <a>DID method's</a> "Resolve" operation that a [=DID resolver=] supports. 
+		</p>
 
 	<p>Note that while this specification defines some base-level functionality for DID resolution, the actual steps
 	required to communicate with a DID's <a>verifiable data registry</a> are defined by the applicable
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1013,63 +1007,73 @@ <h1>DID URL Dereferencing</h1>
...
 		<img style="margin: auto; display: block; width: 75%;"
-			 src="diagrams/section-5/did-url-dereference-overview.svg" alt="
-DIDs resolve to DID documents; DID URLs contains a DID; DID URLs dereferenced to DID document fragments or
-external resources.
+			 src="diagrams/section-5/did-url-dereference-overview-2.svg" alt="
+DIDs URLs resolve to DID documents; DID URLs contains a DID; DID URLs dereferenced to resources.
         " >
 		<figcaption>
-			Overview of DID URL dereference
+			Overview of DID URLk resolution and dereferencing
```
msporny, line 1014:
```suggestion
			Overview of DID URL resolution and dereferencing
```

msporny (approved):

peacekeeper (changes requested):
-1 to changing the input of DID Resolution from a DID to a DID URL. This is an enormous change that would not be backward-compatible with DID 1.0 and break pretty much every existing deployment of DIDs out there.

index.html
```diff
@@ -1013,63 +1007,73 @@ <h1>DID URL Dereferencing</h1>
...
 			See also: <a class="longdesc-link"
 						 href="#did-url-dereference-overview-longdesc">narrative description</a>.
 		</figcaption>
 	</figure>
 
 	<div class="longdesc" id="did-url-dereference-overview-longdesc">
 		<p>
-			The top left part of the diagram contains a rectangle with black outline, labeled "DID".
+			The left part of the diagram contains a single rectangle with
+			black
```
wip-abramson, line 1023:
```suggestion
```

I think

index.html
```diff
@@ -175,17 +174,12 @@
...
 
-	<p>[=DID resolution=] is the process of obtaining a <a>DID document</a> for a given <a>DID</a>. This is one
-	of four required operations that can be performed on any DID ("Read"; the other ones being "Create", "Update",
-	and "Deactivate"). The details of these operations differ depending on the <a>DID method</a>.
-	Building on top of [=DID resolution=], <a>DID URL dereferencing</a> is the process of retrieving a representation
-	of a resource for a given <a>DID URL</a>. Software and/or hardware that is able to execute these processes is called
-	a <a>DID resolver</a>.
+	<p>[=DID resolution=] is the process of obtaining a <a>DID document</a> for a given <a>DID URL</a> (all <a>DIDs</a> are DID URLs). 
+	Relying on [=DID resolution=], <a>DID URL dereferencing</a> is the process of applying a representation
+	of a resource for a given <a>DID URL</a> to the current computational context. Software and/or hardware that is able to execute [=DID resolution=] is called	a <a>DID Resolver</a>. Software and/or hardware that is able to call [=Resolve=] on a DID resolver as part of dereferencing a DID URL are called DID URL clients.
```
wip-abramson, line 179:
```suggestion
	of a resource for a given <a>DID URL</a> to the current computational context.
	Software and/or hardware that is able to execute [=DID resolution=] is called a <a>DID Resolver</a>. 
	Software and/or hardware that is able to call [=Resolve=] on a DID resolver as part of 
	dereferencing a DID URL are called DID URL clients.
```

Split into lines.

Also, I am unsure if [=Resolve=] is intended as a reference to a term. Currently, there is no Resolve term defined in `terms.html`.

index.html
```diff
@@ -894,16 +888,16 @@ <h3>DID Document Metadata</h3>
...
 		<p>A <a>DID resolver</a> implements the following <a>DID resolution</a> algorithm.</p>
 		<ol class="algorithm">
-			<li>Validate that the <var>input <a>DID</a></var> conforms to the `did` rule of the
-			<a href="https://www.w3.org/TR/did-core/#did-syntax">DID Syntax</a>.
+			<li>Validate that the <var>input <a>DID URL</a></var> conforms to the `did-url` rule of the
+			<a href="https://www.w3.org/TR/did-core/#did-url-syntax">DID URL Syntax</a>.
 			If not, the <a>DID resolver</a> MUST return the following result:
 			<ol class="algorithm">
-				<li><b>didResolutionMetadata</b>: <var>error object</var> with type set to <code><a href="#INVALID_DID">https://www.w3.org/ns/did#INVALID_DID</a></code></li>
+				<li><b>didResolutionMetadata</b>: <var>error object</var> with type set to <code><a href="#INVALID_DID_URL">https://www.w3.org/ns/did#INVALID_DID_URL</a></code></li>
```
wip-abramson, line 895: Noting this collapses two different error states into one error code.

1. The DID itself is invalid per either the DID core or DID method spec
2. The URL part of the DID is invalid, e.g. query params are wrong time

I understand that a DID is a DID URL, but still these do fee different enough to distinguish to me.

If we do want to just use INVALID_DID_URL, then this PR should also remove the INVALID_DID error code as it is no longer used by the spec.

wip-abramson:
Chair hat off - I am fine with this change, I think it simplifies the API and makes sense to me. 

I like that with this change - Resolve `did:example:abc?versionId=4` is now the correct language. 

Chair hat on.

I hear Markus's point:

> This is an enormous change that would not be backward-compatible with DID 1.0 and break pretty much every existing deployment of DIDs out there.

There are normative statements in DID 1.0 that define the Resolve API - https://www.w3.org/TR/did-1.0/#resolution. This change is breaking. Are there implications in relation to our charter - with DID being in maintenance mode - that we need to consider here?

Also @peacekeeper will you FO if this PR goes into the spec?

swcurran:
I agree with this change and think it is worth effort to understand the backwards compatibility and try to make the change palatable.  Whether we say it in the spec or not, I think it is what happens so addressing it sooner than later is a good thing.

peacekeeper:
> Also @peacekeeper will you FO if this PR goes into the spec?

Yes, especially since I haven't heard any explanation for this change that made sense to me...

peacekeeper:
@jandrieu It seems the updated image `diagrams/section-5/did-url-dereference-overview-2.svg` isn't included in the PR, can we see this anywhere? It might help understand your perspective.

peacekeeper:
I recognize there was a vote on this, but just for reference, here are the slides I used to argue against this change (I doubt that all the implications have been considered before the vote): https://lists.w3.org/Archives/Public/public-did-wg/2026Jun/0003.html

And I'm not a W3C process expert, but I suspect that this change violates the DID WG charter, and that the PING horizontal review would have to be re-done.

index.html
peacekeeper: I agree, if this change is made, there will be no more use of the INVALID_DID error, since there wouldn't be any more function or algorithm or operation that deals simply with DIDs. Everything will always deal with full DID URLs everywhere. This is also the point I was trying to make in https://github.com/w3c/did-resolution/issues/336#issuecomment-4565150973.

peacekeeper:
Can we close this, or do people still think we should be "resolving DID URLs"?

PR #326, opened 2026-04-12
Andrieu/remove dereferencing function

jandrieu:
This is an attempt to change just the minimum to remove the derefencing function.

This includes 
1. Renamed "dereferencer" to "DID URL client"
1. Refactored the algorithm 
     a.  support the explicit step of determining the retrieval strategy based on the DID document.
     b.  create coherent, prioritized heuristic for determining the retrieval strategy, based on the DID URL and the DID document. This explicitly suggests a hierarchy from Service \-\> serviceType \-\> property \-\> metadata \-\> did method specific \-\> just return DID document
    c. add the final step of using the resource, which is really the business purpose defined by the dereferencer

1. Removed network-based dereferencing
1. Removed dereferencing the fragment section as this is handled by "use the resource" section which defers to the media type of the resource.
1. Removed DID URL Dereferencing Result
1. Removed the dereferencing API and its HTTPS binding to reframe dereferencing as an algorithm to be executed by any client that wants to use a DID URL.
1. Separated the algorithms for service and service type for simplicity.
1. Removed the `accept` options for dereferencing, since there is no longer an http endpoint to receive it. 
1. Added the step where the dereferencer selects a resolver for a given DID method.

***
<a href="https://pr-preview.s3.amazonaws.com/LegReq/did-resolution/pull/326.html" title="Last updated on Apr 13, 2026, 1:27 AM UTC (5a27b90)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-resolution/326/ea94ecc...LegReq:5a27b90.html" title="Last updated on Apr 13, 2026, 1:27 AM UTC (5a27b90)">Diff</a>

PR #331, opened 2026-05-12
Draft refactored dereferencing algorithm outline

wip-abramson:
Okay, here is an initial attempt at the refactored algorithm outline based on the discussion in the WG call. Noting that we did not quite get through all of it.

My sense from the discussion is:

Steps 1 prepare to resolve and steps 2 execute resolution request have broad consensus. 
Modulo the change from DID to DID URL as an input to a resolution request that I know at least @peacekeeper disagrees with. Whilst I do want to get to a WG decision on this matter and move on, I have tried to separate this issue from the high level refactor this PR proposes.

Step 3 Determine Retrieval Strategy also felt like we had good agreement on as a high level step. @peacekeeper noting that you have mentioned you do not like term "Retrieval Strategy". We could change the name and I am open to proposals here. Although dereferencing does involve retrieving a representation of a resource - I dont think its a bad name.

Step 3 involves handling the path and query components of a DID URL. We are going to need to dive into the details of step 3 later on - hopefully Thursday - to unpack what this actually means.

Step 4 Retrieve the resource. This is where #326 diverges with the current spec text most clearly. The current spec does not attempt to retrieve the resource at all, it just returns the URLs to the resource and leaves retrieval out of scope. As noted this implies that retrieval involved HTTP GET on a URL, which is not always the case. It does not create space for extensions to define custom retrieval strategies for their use case. We noted that all services should define how to retrieve the resource identified by the serviceEndpoint, rather than leaving it implicit.

We also discussed, and agree, that it is not the DID Resolution specs responsibility to define how to retrieve resources of different types. The client either knows how to retrieve a resource according to some strategy and can do so, or they do not. I tried to capture that in Step 4.

Step 5 Use the Resource. We ran out of time for this one, and I think it is the step that has the least consensus. At least I know the name is throwing some people off. This step is not really defining how someone should use a resource, but that the resource is available within the application. It also handles the fragment processing part of dereferencing if present. 

I tried to capture that Step 5 may either return the resource or use it within the application context.

I haven't yet specified precisely the inputs and outputs of the different steps, although I think the text does make it pretty clear. 
I also haven't wrote a new intro to the DID URL Dereferencing section or attempted to define inputs and (optional) outputs to the algorithm as a whole.

There are also some terms that @jandrieu used in #326 that I think would be useful to include that I have not yet included. E.g. 
- DID URL Client - currently I just use client.
- Base DID URL - This is the DID URL without the fragment.

Happy to add these in, but wanted to get a sense check from the group. We can discuss on Wednesday.

***
<a href="https://pr-preview.s3.amazonaws.com/wip-abramson/did-resolution/pull/331.html" title="Last updated on May 22, 2026, 2:19 PM UTC (11aac14)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-resolution/331/3a0af5f...wip-abramson:11aac14.html" title="Last updated on May 22, 2026, 2:19 PM UTC (11aac14)">Diff</a>


peacekeeper (changes requested):
> change from DID to DID URL as an input to a resolution request

I'm strongly -1 to this change. I think it would break A LOT of existing DID implementations and infrastructure and would not be backward compatible.

It's completely incomprehensible to me why people want to make this change??

We should keep the clear separation between "resolving a DID" and "dereferencing a DID URL" that we have always had.

wip-abramson:
Yes @peacekeeper I am aware you are strongly against this change.

This PR is not about the change from DID to DID URL as the input to DID Resolution. We are trying to get consensus on the high level outline for refactoring the dereferencing algorithm so things can start moving forwards again. Can you review and provide feedback on this outline please.

I do want to have time on a call to discuss and make a final decision on DID vs DID URL as DID Resolution input. I think there are good arguments for changing to DID URL, e.g. the client does not need to understand the query parameters that are method specific or how to parse them. I also think there are ways to make this change while still being backwards compatible. A DID is a kind of DID URL after all.

As you have been unable to join calls recently, we may end up having to make this decision without you. I would prefer to discuss it first.

ottomorac (approved):

jandrieu:
+1 to merging this.

For the record, I think it is a mistake to advance the spec in the manner the group has decided, as it leaves the spec in an inconsistent state.

Hopefully we can resolve quickly that the public drafts are coherent again.

peacekeeper:
> DID vs DID URL as DID Resolution

I think this change would make most DID 1.0 resolvers non-conformant, because they expect to resolve a DID, not a DID URL. If you pass a DID URL with a path and query into a DID 1.0 resolver, I'm pretty sure that most of them will break. Same for most of the DID methods.

> This PR is not about the change from DID to DID URL as the input to DID Resolution.

Could you then maybe remove the language from the PR that talks about "resolving DID URLs" so we can review just the proposed new outline without such substantive changes?

wip-abramson:
I thought I had already done that. If you get a chance could you propose changes

index.html
```diff
@@ -1299,10 +1299,81 @@ <h2>DID Resolution Algorithm</h2>
...
-
-    <section id="dereferencing">
+		<section id="dereferencing">
       <h1>DID URL Dereferencing</h1>
 
+			<section id="dereferencing-algorithm">
+				<h1>DID URL Dereferencing Algorithm</h1>
+
+				<section>
+					<h1>Prepare to resolve DID URL</h1>
```
```suggestion
					<h1>Prepare for resolution</h1>
```

Maybe this @peacekeeper

index.html
```diff
@@ -1299,10 +1299,81 @@ <h2>DID Resolution Algorithm</h2>
...
+
+					<p>
+The first step in dereferencing a [=DID URL=] is to prepare the DID URL for
+resolution. This involves validating the DID URL syntax, selecting an
+appropriate [=DID Resolver=] and preparing the DID URL and any accompanying
```
```suggestion
appropriate [=DID Resolver=], and preparing the DID URL and any accompanying
```

index.html
```diff
@@ -1299,10 +1299,81 @@ <h2>DID Resolution Algorithm</h2>
...
+					</p>
+
+				</section>
+
+				<section>
+					<h1>Execute Resolution Request</h1>
+					<p>
+The next step is to use the selected [=DID Resolver=] to execute a resolution
+request using the prepared inputs. This may involve making a network request
+to an external service, or it may involve a local resolution process,
```
```suggestion
request using the prepared inputs. This could involve making a network request
to an external service, or it could involve a local resolution process,
```

index.html
```diff
@@ -1299,10 +1299,81 @@ <h2>DID Resolution Algorithm</h2>
...
+
+
+				</section>
+				<section>
+					<h1>Retrieve the Resource</h1>
+          <p>
+The next step is to retrieve the resource identified by the DID URL using the
+determined retrieval strategy. Retrieval strategies may be defined in external
+specifications. It is up to the client to determine if they are able to retrieve
+the resource using the specified strategy.
```
```suggestion
The next step is to retrieve the resource identified by the DID URL, using the
determined retrieval strategy. Retrieval strategies may be defined in external
specifications. It is up to the client to determine whether they are able to
retrieve the resource using the specified strategy.
```

index.html
```diff
@@ -1299,10 +1299,81 @@ <h2>DID Resolution Algorithm</h2>
...
+          </p>
+
+				</section>
+
+				<section>
+					<h1>Use the Resource</h1>
+          <p>
+The final step is to use the retrieved resource to complete the DID URL
+dereferencing process in the context in which dereferencing was invoked.
+This may involve applying a fragment identifier to the retrieved resource,
```
```suggestion
This might involve applying a fragment identifier to the retrieved resource,
```

index.html
wip-abramson: @peacekeeper really looking to get this merged. Would appreciate your input since you are the only one objecting to the contents of this PR

peacekeeper: I'm sorry but this PR still suggests "resolving DID URLs" in several places, this would be a breaking change that I think is really not acceptable...

wip-abramson: Well I would appreciate it if you would propose changes to all the places. I think it is only this one.

The rest are saying prepare DID URL for resolution. That might mean extract the DID and resolve the DID. Or it might mean resolve the DID URL. It is ambiguous at this moment. Intentionally so.

I just want to get the scaffold in for the refactored algorithm so we can move forward to the details of defining the actual algorithm steps. 

The WG needs to discuss DID vs DID URL for resolution. I understand your concern that this is a breaking change. The intention of this PR is not to make or imply that we have made this decision.

wip-abramson:
Hmm, I am confused what has happened here. Now the PR is showing every line has changes against it. It definitely didn't when I submitted it :(

wip-abramson:
Okay. So I fixed the PR although could do with some advice around line endings.

The line endings of main are LF (but maybe I did that in my formatting). My PR also had LR, but when I applied the suggestions, for some reason github changed them to CRLF.

Do we have a preference? Feel like LF is right looking online, so no idea why the code suggestions should have changed this.

Anyway, it is clean now.

@peacekeeper I intend to merge this PR on Monday at the latest. 

I understand you are unhappy about the DID vs DID URL distinction, I have tried to fix that.

When we decide on DID vs DID URL for DID Resolution we can come back in and clarify the text further.

swcurran (approved):

swcurran:
Merging as per the approvals and the guidance of the DID Working Group following discussions about the PR at the last few meetings.  As @jandrieu [notes](https://github.com/w3c/did-resolution/pull/331#issuecomment-4441871906) this PR puts the spec into an inconsistent state and we will need to work as quickly as possible to rectify that via additional PRs.

peacekeeper:
I object to this merge, there was no consensus on this. This PR blurs the lines between DID Resolution and DID URL Dereferencing, which is highly problematic. I request that the PR be reverted.

peacekeeper:
To clarify (once again) the problems I see with this PR:
- "prepare the DID URL for resolution"
- "preparing the DID URL and any accompanying DID Resolution Options for a resolution request"

Saying or suggesting that DID URLs are "resolved" is a major breaking change that is not compatible with DID 1.0, see: https://www.w3.org/TR/did-1.0/#did-resolution

Issue #332, opened 2026-05-14
Experiments with service, serviceType, relativeRef vs. PathService


peacekeeper:
During the [14th May 2026 W3C DID WG meeting](https://www.w3.org/events/meetings/fd89d711-ffa0-4ae1-90bc-b960302dfd3e/20260514T080000/) there were proposals for deprecating the `service`, `serviceType`, and `relativeRef` DID parameters, and there was a claim that the new proposed `PathService` path handling was a superset of those parameters.

So, here’s a challenge. Please explain how this use case could be fulfilled with `PathService` instead of the mentioned DID parameters. Please do this without imposing changes or constraints to the resulting final service endpoint URLs.

Given a DID: `did:ex:1234`

And given this DID document:

```
{
  "@context": [
    "https://www.w3.org/ns/did/v1",
    "https://w3id.org/security/suites/ed25519-2020/v1"
  ],
  "id": "did:ex:1234",
  "verificationMethod": [
    {
      "id": "did:ex:1234#key-1",
      "type": "Ed25519VerificationKey2020",
      "controller": "did:ex:1234",
      "publicKeyMultibase": "z6Mkn7s7ofHS6YMFqqUnevNWmKqcytnnyBRRuopt8YVgEUnf"
    }
  ],
  "authentication": [
    "did:ex:1234#key-1"
  ],
  "assertionMethod": [
    "did:ex:1234#key-1"
  ],
  "service": [
    {
      "id": "did:ex:1234#dpp-1",
      "type": "ProductPassport",
      "serviceEndpoint": "https://example.dpp-provider.com/"
    }
  ]
}
```

I am now constructing two DID URLs:

**DID URL 1:**

```
did:ex:1234?service=dpp-1&relativeRef=%2Fproduct%2F9f7866e1-4bc5-4495-9f75-308350a37280%3Fversion%3Dlatest
```

How this is dereferenced with the current DID URL Dereferencing algorithm:

1. Resolve `did:ex:1234` to the current DID document.
2. Select the service with relative id `#dpp-1`.
3. Apply the relative reference in `relativeRef` to the service endpoint URL `https://service.dpp-provider.com/`

The final service endpoint URL is:

```
https://example.dpp-provider.com/product/9f7866e1-4bc5-4495-9f75-308350a37280?version=latest
```

**DID URL 2:**

```
[...]

PR #335, opened 2026-05-27
Add in the algorithm for handling strategy determination

swcurran:
Replaces the placeholder "Determine Retrieval Strategy" section with a concrete algorithm defining how to select the appropriate handling strategy after DID resolution.

There are 5 conditions checked to arrive at 1 of 4 strategies (DID Doc, DID Method Specific, Object, Path Handling). "Bare DID" (no path or parameters) and "no other conditions apply" both lead to DID Doc.

Per the discussion at the DID Working Group replaced the term "Retrieval Strategy" with "Handling Strategy", but am open to a better term.

Signed-off-by: Stephen Curran <swcurran@gmail.com>

***
<a href="https://pr-preview.s3.amazonaws.com/swcurran/did-resolution/pull/335.html" title="Last updated on May 27, 2026, 6:58 PM UTC (3eba096)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-resolution/335/689679d...swcurran:3eba096.html" title="Last updated on May 27, 2026, 6:58 PM UTC (3eba096)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1343,25 +1343,112 @@ DID document and associated metadata, or error information.
 				</section>
 
 				<section>
-					<h1>Determine Retrieval Strategy</h1>
+					<h1>Determine Handling Strategy</h1>
 					<p>
-After successfully resolving the DID document and accompanying metadata,
-the next step is to determine the appropriate strategy for retrieving
-the resource identified by the [=DID URL=]. This involves analyzing the
-contents of the [=DID resolution=] result, including the DID document and
-its metadata along with the path and query components of the
-DID URL, to determine the method for retrieving the resource.
-				</p>
-
-
+After successfully resolving the [=DID=] and obtaining the [=DID document=] and
+accompanying metadata, the next step is to determine the appropriate strategy
+for handling the [=DID URL=]. This involves analyzing the contents of the [=DID
+resolution result=], including the [=DID document=] and [=DID document
+metadata=] along with the path and query components of the [=DID URL=], to
+determine the handling strategy.
+					</p>
+					<p>
+Execute the following steps until a handling strategy is determined.
+					</p>
+					<ol class="algorithm">
+						<li>
+<b>Bare DID</b>: If there is no path and no query parameters in the
+[=DID URL=], set the handling strategy to <b>Return the DID Document</b>.
+						</li>
+						<li>
+<b>DID Method Specific</b>: If the relevant <a>DID method</a> has a [=DID URL=]
+handling process, set the handling strategy to <b>DID Method Specific</b>. For
+example, a <a>DID method</a> based on a specific blockchain that uses a [=DID URL=] to
+reference a resource stored on that blockchain will process the [=DID URL=] in a
+<a>DID method</a>- (or blockchain-) specific way to retrieve the resource.
+						</li>
+						<li>
+<b>Identified Object</b>: If there is a query parameter defined in this
+specification, in the <a>DID method</a> specification, or in a DID Extension
+[[?DID-EXTENSIONS-PROPERTIES]] that identifies a specific object in the [=DID
+document=] or [=DID document metadata=], set the handling strategy to <b>Process
+Object</b>. For example, the
+<code>service</code> query parameter defined in this specification
+<a href="#did-parameters"></a> identifies a specific service object to process.
+If there is also a path in the [=DID URL=], the handling may involve the use of
+the path.
+						</li>
+						<li>
+<b>Path</b>: If the [=DID URL=] contains a path, use the following
+steps to determine the handling strategy.
+							<ol class="algorithm">
+								<li>
+Collect all Path Handling Objects within the [=DID document=] and
+[=DID document metadata=]. A Path Handling Object is defined as any
+JSON object that satisfies both of the following conditions:
+									<ol class="algorithm">
+										<li>
+The object has a <code>type</code> member whose value either:
+											<ul>
+												<li>
+is the string <code>"PathHandlingObject"</code>, or
+												</li>
+												<li>
+is an array that contains the string
+<code>"PathHandlingObject"</code> as one of its elements; and
+												</li>
+											</ul>
+										</li>
+										<li>
+The object has a <code>path</code> member.
+										</li>
+									</ol>
+									<p>
+Objects that do not satisfy both conditions MUST be ignored for
+the purposes of path handling strategy determination.
+									</p>
+								</li>
+								<li>
+If no objects are found, return an error.
+								</li>
+								<li>
+Extract the path from the [=DID URL=] and the value of the
+<code>path</code> member of each collected object, compare the two
+strings from the beginning, character by character, stopping at the
+first mismatch. Record the number of characters matched for each object.
+								</li>
+								<li>
+Select the objects with the largest number of matched characters.
+									<ol class="algorithm">
+										<li>
+If all of the matches are 0, return an error.
+										</li>
+										<li>
+If more than 1 object has the same largest number of matched characters, return
+an error.
+										</li>
+									</ol>
+								</li>
+								<li>
+Set the handling strategy to <b>Resource at Path</b>. The strategy
+will use the selected path handling object and may use parameters that are
+included in the [=DID URL=].
+								</li>
+							</ol>
+						</li>
+						<li>
+<b>Default</b>: If no other strategy has been determined, set the handling strategy
+to <b>Return the DID Document</b>.
+						</li>
+					</ol>
 				</section>
 				<section>
-					<h1>Retrieve the Resource</h1>
+					<h1>Apply the Handling Strategy</h1>
           <p>
-The next step is to retrieve the resource identified by the DID URL, using the
-determined retrieval strategy. Retrieval strategies may be defined in external
-specifications. It is up to the client to determine whether they are able to
-retrieve the resource using the specified strategy.
+The next step is to apply the handling strategy identified by the DID URL.
+Handling strategies may be defined in external specifications. It is up to the
+client to determine whether they are able to handle the DID URL using the
+specified strategy.
           </p>
 
 				</section>


peacekeeper (changes requested):
I like the general direction of this, and I like the concept of "handling strategies", to make the algorithm more modular and extensible. But I think a bit more work is needed before this can be adopted, e.g.:

- It's unclear how `service`, `serviceType`, `relativeRef` will work in this framework. There was a claim that `PathService` would be a superset of those parameters, but I don't think it has been shown how that would work (see https://github.com/w3c/did-resolution/issues/332).
- What if both `service` and `serviceType` are present? In the current algorithm this is supported. Would this be handled by a single handling strategy, or can multiple handling strategies "work together"?
- It's unclear how `relativeRef` would be processed. Would this be used by both the "Identified Object" strategy (if a service is identified) and the "Resource at Path" strategy? We need to consider that there may be parameters and options that are relevant across multiple strategies.
- It's unclear how the `versionId`, `versionTime` parameters will be processed. Would they be covered by the "Return the DID Document" strategy?
- I'm undecided about the idea of "Path Handling Objects". This is yet again a new concept that doesn't exist in the core data model. I feel like it would be better for this to be specified in an extension, not in DID Resolution itself, just like linked resources and DID linked resources are also defined in extensions.
- It's unclear how the `verificationRelationship` option would be processed. Would this be processed as part of a handling strategy, or in a later step?
- It's unclear how extension parameters such as `transformKeys` would be processed.
- I think it's a mistake to make the default strategy to be to return the DID document. In the current algorithm, the default is to return a NOT_FOUND error.

wip-abramson:
> It's unclear how the versionId, versionTime parameters will be processed. Would they be covered by the "Return the DID Document" strategy?

These parameters are not handled by the dereferencing algorithm at all. They are unavoidably part of resolution and need to be passed to the resolver. This is part of the argument for passing a DID URL to the resolver, because the URL can contain resolution specific query parameters.

index.html
```diff
@@ -1343,25 +1343,112 @@ <h1>Execute Resolution Request</h1>
...
-its metadata along with the path and query components of the
-DID URL, to determine the method for retrieving the resource.
-				</p>
-
-
+After successfully resolving the [=DID=] and obtaining the [=DID document=] and
+accompanying metadata, the next step is to determine the appropriate strategy
+for handling the [=DID URL=]. This involves analyzing the contents of the [=DID
+resolution result=], including the [=DID document=] and [=DID document
+metadata=] along with the path and query components of the [=DID URL=], to
```
wip-abramson, line 1352: We should define DID document metadata in the terminology I think

index.html
```diff
@@ -1343,25 +1343,112 @@ <h1>Execute Resolution Request</h1>
...
+example, a <a>DID method</a> based on a specific blockchain that uses a [=DID URL=] to
+reference a resource stored on that blockchain will process the [=DID URL=] in a
+<a>DID method</a>- (or blockchain-) specific way to retrieve the resource.
+						</li>
+						<li>
+<b>Identified Object</b>: If there is a query parameter defined in this
+specification, in the <a>DID method</a> specification, or in a DID Extension
+[[?DID-EXTENSIONS-PROPERTIES]] that identifies a specific object in the [=DID
+document=] or [=DID document metadata=], set the handling strategy to <b>Process
+Object</b>. For example, the
```
wip-abramson, line 1375: I don't think the handling strategy is Process Object. Probably something like

"determine the handling strategy from the type of the object"

wip-abramson:
> It's unclear how extension parameters such as transformKeys would be processed.

This is a good point. We definitely need to think about when extension parameters are processed. My sense if the client should process all options they understand, but in which order? And if they don't understand a parameter should they just ignore it?

peacekeeper:
> It's unclear how the versionId, versionTime parameters will be processed. Would they be covered by the "Return the DID Document" strategy?

> These parameters are not handled by the dereferencing algorithm at all. They are unavoidably part of resolution and need to be passed to the resolver.

Yes they are part of resolution, but if you have a DID URL like `did:ex:123?versionId=1`, it wouldn't trigger the first handling strategy in this PR, i.e. the "Bare DID" one. Actually, now that I think about it, this is a problem that also exists in the algorithm I wrote.

peacekeeper:
> My sense if the client should process all options they understand, but in which order? And if they don't understand a parameter should they just ignore it?

I thought that on today's DID WG call, there was consensus that the client shouldn't have to understand parameters and options, and just pass them all to the resolver?

I guess extensions like [transformKeys](https://github.com/decentralized-identity/did-spec-extensions/blob/main/parameters/transform-keys.md) or [expandReferencedVerificationMethods](https://github.com/w3c/did-extensions/issues/621) or [returnDeactivatedDidDocument](https://github.com/w3c/did-extensions/issues/622) would be processed by resolvers that support them, and ignored by others. Theoretically, extensions like this could also be processed by the client, but it would require clients to distinguish which parameters and options it can process. In any case it's not clear to me how extensions like this fit into the structure introduced by this PR.

wip-abramson:
> I thought that on today's DID WG call, there was consensus that the client shouldn't have to understand parameters and options, and just pass them all to the resolver?

They shouldn't have to understand and parse parameters that they don't recognise and understand. But some query parameters are specifically for dereferencing, clients dereferencing DID URLs should process all parameters they know how to process.

> I guess extensions like [transformKeys](https://github.com/decentralized-identity/did-spec-extensions/blob/main/parameters/transform-keys.md) or https://github.com/w3c/did-extensions/issues/621 or https://github.com/w3c/did-extensions/issues/622 would be processed by resolvers that support them, and ignored by others.

I do not think resolvers should be doing any transformation to the DID document. Resolution should return the canonical DID document determined by the DID methods resolve operation. These parameters should be treated as dereferencing parameters IMO and clients that understand them can choose to process them.

Return Deactivated DID Document does make sense to me as a parameter handled by the resolver though

wip-abramson:
> Yes they are part of resolution, but if you have a DID URL like `did:ex:123?versionId=1`, it wouldn't trigger the first handling strategy in this PR, i.e. the "Bare DID" one. Actually, now that I think about it, this is a problem that also exists in the algorithm I wrote.

This is a good point. 

> Default: If no other strategy has been determined, set the handling strategy to Return the DID Document.

I think you could remove step 1 in this algorithm and just have the default step, if no other retrieval strategy identified be to return the DID document.

I saw you said:

> I think it's a mistake to make the default strategy to be to return the DID document. In the current algorithm, the default is to return a NOT_FOUND error.

But I think this cannot be right, because the DID URL can contain parameters that are targeting resolution e.g. versionId, but also any extensions. The client can't know if they are for dereferencing or resolution. 

So I think the algorithm should be:

1. Process all query parameters (what order?) understood by the dereferencing client
2. Handle path if present
3. If no other retrieval strategy identified, return DID document

Some query parameters do apply transformations to the DID document e.g. `transformKeys`. Others target specific retrieval strategies e.g. `service`

What happens if both kinds present. e.g. `did:ex:123?transformKeys=JWK&service=my-service`

I think the algorithm above works, transformKeys just becomes irrelevant. Maybe processing order of query parameters doesn't matter. Instead it is all about the identified retrieval strategy after processing all parameters. This is either a single identified retrieval strategy, multiple strategies so an error, or fall back to return DID document.

e.g. there could be an extension `linkedResource` that works like service but for `linkedResource` in that case a DID URL like this: `did:example:123?linkedResource=my-resource&service=my-service`.

Would identify multiple retrieval strategies and hence would fail. Note, there is no ordering here - service does not superseed linkedResource. However, say the client does not understand the linkedResource query parameter I think then would this simply be ignored and the service would be targeted?

Does this help us with `serviceType` too? I know we said we would push to an extension, and that still makes sense but we should understand how it would work. I think the approach outlined means if multiple services of the same type in the DID document then throw an error.

I do think is a query parameter like `service` is present but the targeted service `my-service` is not in the DID document then a NOT_FOUND error makes sense.

index.html
swcurran: That seems meta. The purpose of this step is to determine the handling strategy, so the output shouldn't be "determine handling strategy". :-)   I'm not thrilled with the name, but there you go.  Open to ideas on it.

peacekeeper:
> I do not think resolvers should be doing any transformation to the DID document. Resolution should return the canonical DID document determined by the DID methods resolve operation. These parameters should be treated as dereferencing parameters IMO and clients that understand them can choose to process them.

That makes sense to me, but I think it completely contradicts what I have been hearing on the last few WG calls, where some people argued that clients shouldn't parse or process parameters in any way.

Personally, I think there is nothing wrong with the resolver processing transformKeys or expandReferencedVerificationMethods or returnDeactivatedDidDocument. Those should all work simply as options passed to the resolver, without ever having to construct a DID URL with them encoded as DID parameters.

swcurran:
Some notes on @peacekeeper 's [comment here](https://github.com/w3c/did-resolution/pull/335#pullrequestreview-4380579803).

Something to keep in mind from both @wip-abramson and @peacekeeper comments.  This PR ONLY covers step 3 (determine strategy) of @wip-abramson 's process.  Step 3 is after DID Resolution is complete (we have the DIDDoc and DIDDoc Metadata) and before we do anything (execute the handling strategy).

> It's unclear how service, serviceType, relativeRef will work in this framework. There was a claim that PathService would be a superset of those parameters, but I don't think it has been shown how that would work (see https://github.com/w3c/did-resolution/issues/332).

I deliberately only included `service` because that was the last direction from the Working Group. `serviceType` is to be considered an extension.  `relativeRef` is not important for determining the handling strategy -- but might come into play in executing it.  Likewise, `PathService` is relevant only in executing the selected strategy.

> What if both `service` and `serviceType` are present? In the current algorithm this is supported. Would this be handled by a single handling strategy, or can multiple handling strategies "work together"?

My read of the leaning of the Working Group was to not allow that (or multiple `service` parameters).  We will need to formally decide that — I deliberately didn’t address it, but plan to discuss it.

It is tricky to allow room for supporting any possible DID Extension that people can think up. I’m hoping that is part of the discussion related to this PR and following ones. Further comments on that later in this comment.

> It's unclear how `relativeRef` would be processed. Would this be used by both the "Identified Object" strategy (if a service is identified) and the "Resource at Path" strategy? We need to consider that there may be parameters and options that are relevant across multiple strategies.

Yes — used by both. And as noted above — how it used is out of scope for this PR.

> It's unclear how the `versionId`, `versionTime` parameters will be processed. Would they be covered by the "Return the DID Document" strategy?

They will be processed by the DID Resolution step already completed by the time we get here, and thus are ignored in this step. The fact that there are “Resolution Parameters” and “Dereferencing Parameters” is an elephant in the room in IMHO. Especially when combined with writing a spec that has to account for “Extensions” that will grow after the spec is defined. At this step, I think they should be silently ignored.

> I'm undecided about the idea of "Path Handling Objects". This is yet again a new concept that doesn't exist in the core data model. I feel like it would be better for this to be specified in an extension, not in DID Resolution itself, just like linked resources and DID linked resources are also defined in extensions.

The core model ignores paths entirely which IMHO is a major missing feature — as we have discussed often. Flagging Path Handling Objects is not an extension, it is a way to enable a class of extensions that will handle paths consistently, including a “standard” way that might avoid needing to have extensions at all for many use cases.

> It's unclear how the `verificationRelationship` option would be processed. Would this be processed as part of a handling strategy, or in a later step?

I looked in the current spec for `verificationRelationship`, and I’m not totally clear on its purpose and haven’t been able to find examples — so let’s discuss. Based on what I could find in search, I suspect it is not used in determining a strategy, but we’ll see.

> It's unclear how extension parameters such as `transformKeys` would be processed.

This property would be used at strategy execution rather than at determination time. I think it would only apply when the "Return DID Doc" strategy was used as its point seems to be to alter the DIDDoc between retrieval from source and returning to the client of the dereferencer.

<rant>On a separate note, I question whether it is an extension that should be considered in scope for DID Resolution. It seems to me it is more of a convenience function a library might use after DID URL Dereferencing. It is this kind of DID extension that I worry about trying to account for in the DID Resolution spec — finding a place for all such “arbitrary” features. My guess is that the vast majority of DID URL Dereferencer implementations will not know how to process this, and then we have to decide if they report an error or silently ignored the parameter.  But of course saying it is "not allowed" creates a governance problem — who decides on what is a valid extension?  </rant>

> I think it's a mistake to make the default strategy to be to return the DID document. In the current algorithm, the default is to return a NOT_FOUND error.

I suggest we look at examples to see what to use. Currently I’ve proposed the “bare DID” use case returns the DID Document as the first step (if only to get it out of the way).  DID URLs when there are parameters that do not impact handle strategy determination (verificationRelationship, transformKeys are examples) are left after other strategies are selected, which is why the fall through is "Return DID Doc".

index.html
wip-abramson: No, the output is not "determine handling strategy".

The output is the handling strategy as determined and defined by the type of the object.

E.g. PathService, LinkedResource or whatever.

These all should define an algorithm for handling paths. This is the handling strategy. 

Step 4, is just execute that algorithm I thought

wip-abramson:
Are there three kinds of query parameters a DID URL could contain?

1. DID Resolution parameters: `versionId`, `versionTime`
2. Target Path Handling Object: `service`, `serviceType`
3. Strategy specific: `transformKeys`, `verificationRelationship`, `relativeRef`?

The dereferencing client may or may not understand how to apply any of the query parameters they encounter in a DID URL:

- For DID Resolution parameters they do not need to understand them, the resolver handles this. But, they cannot know all query parameters that are DID Resolution parameters due to extensions.
- Query parameters known to target a path handling object are applied in Step 3.
- Strategy specific parameters are applied in step 4 by the strategies that are understood to use them by the client. 

So step 3 should include something about apply all understood extension query parameters that are for targetting path handling objects.

I think for all steps, we just pass in the DID URL and the step parses the query parameters it understands for that step.

index.html
swcurran: OK -- I think we are saying roughly the same thing.  I'll check the wording.  At the point we've reached in the big picture algorithm, we've determined not only that we are going to be "processing an object", but the precise object that is going to processed, including its types.  I had assumed that the follow on "process object" step would begin with "If the object type is ...".  I think you are saying that we should do that determination here and have a "handling strategy" for each.  I'm OK with either.

The questions around that (either way) is that the current plan is to have only one published way in this spec that is concrete (PathService or whatever we call it), and then a fallback for any other type that is "Do whatever the extension specification requires".  In other words, I don't think we plan to move LRs or DLRs into this spec.  Is that how you see it?

wip-abramson: > In other words, I don't think we plan to move LRs or DLRs into this spec. Is that how you see it?

Yep

swcurran:
I went through the "DID Extensions Registries" and we might want add a 4th group to @wip-abramson list in the previous comment -- "Result Parameters", parameters that are applied at the end of the process, regardless of the handling. Included from the registry are "accept" and "hl".

The extensions also include "initialState" and "signedIetfJsonPatch" which I think goin in the DID Resolution parameters.  There is also "resource", which I think target path handling, but it also says it is "being replaced by DID Linked Resources" so can probably be ignored.

Agree that each step gets the full DID URL, but processes only the elements it understands. That raises the tricky question of ignored items.  Do we need a global view of elements that were ignored by all steps?

wip-abramson:
> That raises the tricky question of ignored items. Do we need a global view of elements that were ignored by all steps?

I don't think this is possible, since dereferencing clients cannot know if parameters they have ignored were applied by the DID resolver.

index.html
```diff
@@ -1343,25 +1343,112 @@ <h1>Execute Resolution Request</h1>
...
+metadata=] along with the path and query components of the [=DID URL=], to
+determine the handling strategy.
+					</p>
+					<p>
+Execute the following steps until a handling strategy is determined.
+					</p>
+					<ol class="algorithm">
+						<li>
+<b>Bare DID</b>: If there is no path and no query parameters in the
+[=DID URL=], set the handling strategy to <b>Return the DID Document</b>.
```
jandrieu, line 1361: This is incorrect for versionId and versionTime, and basically wrong for any query term that isn't a service invocation or path handling call.

swcurran: Based on the need to include "known extension properties", I agree this step is to be removed, but I don't understand your comment.  Could you explain this comment so I can try to understand better your thinking.  The PR text says that by definition of "Bare DID", there is no "versionId" and "versionTime" or other query term.

swcurran:
PR #340 is a replacement of this PR that includes feedback here and discussed in the DID Working Group Special Topic call of 2026.06.10. Closing this PR in favour of that PR.

Please note that some of the feedback in this PR may be relevant to #340 .

PR #339, opened 2026-06-10
Correction to CSS Style to not put algorithm numbering on unordered lists

swcurran:
This is an issue that came up in the initial Path Handling PR that I created (#316 ).  In the current style for the "algorithm" class, when there is an unordered list within the algorithm, it is rendered as both a bullet and numbering.  Searching I found this fix that @pchampin confirmed was correct.  However, since that PR is destined to the scrap heap, I'd like to get the change into the document.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/did-resolution/pull/339.html" title="Last updated on Jun 10, 2026, 6:39 PM UTC (d9473eb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-resolution/339/d008883...d9473eb.html" title="Last updated on Jun 10, 2026, 6:39 PM UTC (d9473eb)">Diff</a>

PR #340, opened 2026-06-10
Add in the algorithm for handling strategy determination - Try 2

swcurran:
This PR is a replacement for #335 that I will close shortly, and reflects the feedback received in #335 and as discussed at the DID Working Group Special Topic call on 2026.06.10.

Notes/Remaining questions:

- This does not included the planned note higher up in the spec about the categories of DID URL query parameters that apply at different stages of the DID URL Dereferencing process. There is a mention of that in this PR.
- I debated whether to do the "DID Object" (@jandrieu 's DIDDoc properties and DIDDoc Metadata properties combined) as before or after the Path handling and I think it is more appropriate after, so that "old-style" objects that have been updated to be DID Handling Objects don't get processed before the DID Handling Objects step. Other opinions?
- I wasn't sure how much normative language was needed so was light on it.
- I've identified the specific errors to be raised, but am not certain I have the right style for that vs. the rest of the document.
- The `[=DID Document Metadata=]` is not a definition, but I think we should add that vs. linking to a section of the spec.  What do others think?

On a ReSpec note -- most of the rest of the document has paragraphs starting in column 1 and wrapped, with the HTML tags indented to show the structure, and that is what I have done. However, the sections around this section do not use that style -- indenting the paragraphs with the tags.  We should be consistent on that, right?

Also, this PR duplicates the change in PR #339 that I submitted earlier today about bullets in the midst of algorithm ordered lists.

Please note that some of the comments of #335 may still apply to this. I'll try to go through them and add a comment about the ones that I see as still relevant.

Signed-off-by: Stephen Curran <swcurran@gmail.com>

***
<a href="https://pr-preview.s3.amazonaws.com/swcurran/did-resolution/pull/340.html" title="Last updated on Jun 16, 2026, 7:46 PM UTC (2407c5e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-resolution/340/d008883...swcurran:2407c5e.html" title="Last updated on Jun 16, 2026, 7:46 PM UTC (2407c5e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -117,7 +117,7 @@
       ol.algorithm li {
         margin: 0.5em 0;
       }
-      ol.algorithm li:before {
+      ol.algorithm > li:before {
         font-weight: bold;
         counter-increment: numsection;
         content: counters(numsection, ".") ") ";
@@ -2178,19 +2178,133 @@ Dereferencing a DID URL to a service endpoint URL.
 
             </section>
 
-            <section>
-                <h1>Determine Retrieval Strategy</h1>
+				<section>
+					<h1>Determine Handling Strategy</h1>
+					<p>
+After successfully resolving the [=DID=] and obtaining the [=DID document=] and
+accompanying metadata, the next step is to determine the appropriate strategy
+for handling the [=DID URL=]. This involves analyzing the contents of the [=DID
+resolution result=], including the [=DID document=] and <a
+href="#did-resolution-metadata">DID resolution metadata</a> along with the path
+and query components of the [=DID URL=], to determine the handling strategy.
+					</p>
+					<p>
+Execute the following steps until a handling strategy is determined. While the
+full [=DID URL=] is available in these steps, query parameters that are not
+applicable to the determination of the handling strategy are ignored.
+					</p>
+					<ol class="algorithm">
+						<li>
+<b>Identified Object</b>: If there are [=DID URL=] query parameters defined in
+this specification, in the <a>DID method</a> specification, or in a DID
+Extension [[?DID-EXTENSIONS-PROPERTIES]] that identifies a specific object in
+the [=DID document=] or [=DID document metadata=], set the handling strategy to
+<b>Process Object</b>. The <code>service</code> query parameter defined in this
+specification <a href="#did-parameters"></a> identifies a specific `service`
+object. If there is also a path in the [=DID URL=], the handling may involve the
+use of the path.
+						</li>
+						<li>
+<b>Path</b>: If the [=DID URL=] contains a path, use the following sub-steps to
+determine the handling strategy.
+							<ol class="algorithm">
+								<li>
+Collect all Path Handling Objects within the [=DID document=] and [=DID document
+metadata=]. A "Path Handling Object" is defined as any JSON object that
+satisfies both of the following conditions:
+									<ul>
+										<li>
+The object has a <code>type</code> member whose value either:
+											<ul>
+												<li>
+is the string <code>"PathHandlingObject"</code>, or
+												</li>
+												<li>
+is an array that contains the string
+<code>"PathHandlingObject"</code> as one of its elements; and
+												</li>
+											</ul>
+										</li>
+										<li>
+The object has a <code>path</code> member.
+										</li>
+									</ul>
+									<p>
+Objects that do not satisfy both conditions MUST be ignored for the purposes of
+path handling strategy determination.
+									</p>
+								</li>
+								<li>
+If no Path Handling Objects are found, skip the remaining sub-steps and continue
+the Determine Handling Strategy algorithm.
+								</li>
+								<li>
+Extract the path component from the [=DID URL=] and the value of the
+<code>path</code> member from each collected Path Handling Object.
+								</li>
+								<li>
+For each collected Path Handling Object, compare the [=DID URL=] path against
+the object's path value character by character from the beginning, stopping at
+the first mismatch or the end of either string. Record the number of matched
+characters for each object.
+								</li>
+								<li>
+Select the objects with the largest number of matched characters.
+									<ul>
+										<li>
+If all of the matches are 0, the error `invalidDIDURL` MUST be returned.
+										</li>
+										<li>
+If more than 1 object has the same largest number of matched characters, the
+error `invalidDIDDoc` MUST be returned.
+										</li>
+									</ul>
+								</li>
+								<li>
+Set the handling strategy to <b>Process Object</b>. The strategy MUST use the
+selected Path Handling Object, the [=DID URL=], the [=DID document=], and MAY
+use the [=DID document metadata=].
+								</li>
+							</ol>
+						</li>
+						<li>
+<b>DID Object</b>: Collect from the [=DID document=] and [=DID document
+metadata=] for known objects that affect dereferencing. If found, use the
+"Process Object" handling strategy based on the [=DID URL=] and the identified
+object(s). For example, linkedResource object uses its own retrieval strategy as
+described at Linked Resource in the [DID-EXTENSIONS-PROPERTIES] specification.
+If there are multiple different objects collected, stop further processing and
+the error `invalidDIDDoc` MUST be returned.
+            <div class="note">
                 <p>
-                    After successfully resolving the DID document and accompanying metadata,
-                    the next step is to determine the appropriate strategy for retrieving
-                    the resource identified by the [=DID URL=]. This involves analyzing the
-                    contents of the [=DID resolution=] result, including the DID document and
-                    its metadata, along with the path and query components of the
-                    DID URL, to determine the method for retrieving the resource.
+This process is intended primarily to cover extensions that predate the "Path
+Handling Object" concept. The "Path Handling Object" approach or a query
+parameter to identify a specific object **SHOULD** be used in place of the
+"existence of the object" mechanism described here.
                 </p>
+            </div>
+						</li>
+						<li>
+<b>DID Method Specific</b>: If the relevant <a>DID method</a> has a [=DID URL=]
+handling process, set the handling strategy to <b>DID Method Specific</b>. For
+example, a <a>DID method</a> based on a specific blockchain that uses a [=DID
+URL=] to reference a resource stored on that blockchain will process the [=DID
+URL=] in a <a>DID method</a>- (or blockchain-) specific way to retrieve the
+resource.
+						</li>
+						<li>
+<b>Default</b>: If no other strategy has been determined, set the handling
+strategy to <b>Return the DID Document</b>.
+						</li>
+					</ol>
+          <p>
+Once the handling strategy is determined, the next step is to retrieve the
+resource using the selected strategy, the [=DID URL=], the [=DID document=], the
+[=DID document metadata=] and the data collected here (notably, the selected
+Path Handling Object or other identified objects) as inputs.
+          </p>
+				</section>
 
-
-            </section>
             <section>
                 <h1>Retrieve the Resource</h1>
                 <p>
--- a/index.html
+++ b/index.html
@@ -2199,10 +2199,10 @@ applicable to the determination of the handling strategy are ignored.
 this specification, in the <a>DID method</a> specification, or in a DID
 Extension [[?DID-EXTENSIONS-PROPERTIES]] that identifies a specific object in
 the [=DID document=] or [=DID document metadata=], set the handling strategy to
-<b>Process Object</b>. The <code>service</code> query parameter defined in this
-specification <a href="#did-parameters"></a> identifies a specific `service`
-object. If there is also a path in the [=DID URL=], the handling may involve the
-use of the path.
+<b>Process Object</b>. The <code>service</code> query parameter defined in
+Section <a href="#did-parameters"></a> identifies a specific `service` object.
+If there is also a path in the [=DID URL=], the handling may involve the use of
+the path.
 						</li>
 						<li>
 <b>Path</b>: If the [=DID URL=] contains a path, use the following sub-steps to
@@ -2214,14 +2214,14 @@ metadata=]. A "Path Handling Object" is defined as any JSON object that
 satisfies both of the following conditions:
 									<ul>
 										<li>
-The object has a <code>type</code> member whose value either:
+The object has a <code>type</code> member whose value is one of the following:
 											<ul>
 												<li>
-is the string <code>"PathHandlingObject"</code>, or
+The string <code>"PathHandlingObject"</code>, or
 												</li>
 												<li>
-is an array that contains the string
-<code>"PathHandlingObject"</code> as one of its elements; and
+An array that contains the string <code>"PathHandlingObject"</code> as one
+of its elements; and
 												</li>
 											</ul>
 										</li>
@@ -2230,8 +2230,8 @@ The object has a <code>path</code> member.
 										</li>
 									</ul>
 									<p>
-Objects that do not satisfy both conditions MUST be ignored for the purposes of
-path handling strategy determination.
+Objects that do not satisfy both conditions MUST be ignored for purposes of
+determining the path-handling strategy.
 									</p>
 								</li>
 								<li>
@@ -2244,7 +2244,7 @@ Extract the path component from the [=DID URL=] and the value of the
 								</li>
 								<li>
 For each collected Path Handling Object, compare the [=DID URL=] path against
-the object's path value character by character from the beginning, stopping at
+the object's path value, character by character, from the beginning, stopping at
 the first mismatch or the end of either string. Record the number of matched
 characters for each object.
 								</li>
@@ -2252,29 +2252,33 @@ characters for each object.
 Select the objects with the largest number of matched characters.
 									<ul>
 										<li>
-If all of the matches are 0, the error `invalidDIDURL` MUST be returned.
+If all of the matches are 0, the error <code><a
+href="#INVALID_DID_URL">https://www.w3.org/ns/did#INVALID_DID_URL</a></code>
+MUST be returned.
 										</li>
 										<li>
 If more than 1 object has the same largest number of matched characters, the
-error `invalidDIDDoc` MUST be returned.
+error <code><a
+href="#INVALID_DID_DOCUMENT">https://www.w3.org/ns/did#INVALID_DID_DOCUMENT</a></code>
+MUST be returned.
 										</li>
 									</ul>
 								</li>
 								<li>
 Set the handling strategy to <b>Process Object</b>. The strategy MUST use the
-selected Path Handling Object, the [=DID URL=], the [=DID document=], and MAY
-use the [=DID document metadata=].
+selected Path Handling Object, the [=DID URL=], and the [=DID document=], and
+MAY use the [=DID document metadata=].
 								</li>
 							</ol>
 						</li>
 						<li>
 <b>DID Object</b>: Collect from the [=DID document=] and [=DID document
-metadata=] for known objects that affect dereferencing. If found, use the
-"Process Object" handling strategy based on the [=DID URL=] and the identified
-object(s). For example, linkedResource object uses its own retrieval strategy as
-described at Linked Resource in the [DID-EXTENSIONS-PROPERTIES] specification.
-If there are multiple different objects collected, stop further processing and
-the error `invalidDIDDoc` MUST be returned.
+metadata=] any objects known to affect dereferencing. If none are found, continue
+to the next step. If any are found, use the specifications that define these
+objects to determine if any are targeted by the DID URL. If more than one is
+targeted, the error <code><a
+href="#INVALID_DID_DOCUMENT">https://www.w3.org/ns/did#INVALID_DID_DOCUMENT</a></code>
+MUST be returned. The targeted object MUST define a handling strategy.
             <div class="note">
                 <p>
 This process is intended primarily to cover extensions that predate the "Path


index.html
```diff
@@ -2178,19 +2178,133 @@ <h1>Execute Resolution Request</h1>
...
+					<ol class="algorithm">
+						<li>
+<b>Identified Object</b>: If there are [=DID URL=] query parameters defined in
+this specification, in the <a>DID method</a> specification, or in a DID
+Extension [[?DID-EXTENSIONS-PROPERTIES]] that identifies a specific object in
+the [=DID document=] or [=DID document metadata=], set the handling strategy to
+<b>Process Object</b>. The <code>service</code> query parameter defined in this
+specification <a href="#did-parameters"></a> identifies a specific `service`
+object. If there is also a path in the [=DID URL=], the handling may involve the
+use of the path.
```
```suggestion
object.
```

Do we need to say this here? I think it will confuse people, because the next step goes into dealing with path. But, that is not the path handling that is meant here.

I think, this is just a note that the handling strategy might use the path. My sense is we can just talk to that in step 4 of the algorithm.

The point is, if you have a query parameter that identifies an object. You use that object and exit this step.

index.html
```diff
@@ -2178,19 +2178,133 @@ <h1>Execute Resolution Request</h1>
...
+applicable to the determination of the handling strategy are ignored.
+					</p>
+					<ol class="algorithm">
+						<li>
+<b>Identified Object</b>: If there are [=DID URL=] query parameters defined in
+this specification, in the <a>DID method</a> specification, or in a DID
+Extension [[?DID-EXTENSIONS-PROPERTIES]] that identifies a specific object in
+the [=DID document=] or [=DID document metadata=], set the handling strategy to
+<b>Process Object</b>. The <code>service</code> query parameter defined in this
+specification <a href="#did-parameters"></a> identifies a specific `service`
```
```suggestion
<b>Process Object</b>. The <code>service</code> query parameter defined in Section <a href="#did-parameters"></a> identifies a specific `service`
```

index.html
```diff
@@ -2178,19 +2178,133 @@ <h1>Execute Resolution Request</h1>
...
+								</li>
+								<li>
+Select the objects with the largest number of matched characters.
+									<ul>
+										<li>
+If all of the matches are 0, the error `invalidDIDURL` MUST be returned.
+										</li>
+										<li>
+If more than 1 object has the same largest number of matched characters, the
+error `invalidDIDDoc` MUST be returned.
```
wip-abramson: These errors should follow https://w3c.github.io/did-resolution/#errors. If we want to return errors here at all. 

I imagine @jandrieu might say something like what is the client returning this error too though.

> If all of the matches are 0, the error `invalidDIDURL` MUST be returned.

I also don't know that this is an error state. Would you not just continue to the legacy fallbacks here.

index.html
```diff
@@ -2178,19 +2178,133 @@ <h1>Execute Resolution Request</h1>
...
+							</ol>
+						</li>
+						<li>
+<b>DID Object</b>: Collect from the [=DID document=] and [=DID document
+metadata=] for known objects that affect dereferencing. If found, use the
+"Process Object" handling strategy based on the [=DID URL=] and the identified
+object(s). For example, linkedResource object uses its own retrieval strategy as
+described at Linked Resource in the [DID-EXTENSIONS-PROPERTIES] specification.
+If there are multiple different objects collected, stop further processing and
+the error `invalidDIDDoc` MUST be returned.
```
wip-abramson: I still don't like "Process Object" as a handling strategy. It seems like it is used as a proxy for the actual handling strategy.

E.g. you are saying the strategy is process this object. But later on a linkedResource object uses its own *handling* strategy.


```suggestion
<b>DID Object</b>: Collect from the [=DID document=] and [=DID document
metadata=] any objects known to affect dereferencing. If none are found, continue to the next step. If found, use the specifications that define these objects to determine if any are targeted by the DID URL. If multiple are targeted, throw an error.  The targeted object MUST define a handling strategy.
```

This is still not quite right. But I am basically trying to say, if you understand objects in the DID Resolution Result that are relevant for dererencing handle them however you understand them. The only rule for the resolution spec here is, if multiple targeted its an error I think. Although maybe even that isn't true/can be changed by extensions?

wip-abramson (commented):
I need to think some more about this. 

The purpose of this step is to identify the handling strategy. And in this text there are 3 types of strategies I think:

- Process Object: Some object defines the strategy
- DID Method specific
- DID Document

Is that right?

Really the purpose of this step is to select an object from the DID Resolution Result that defines how to handle the DID URL. Or if none selected, to return use the DID Document strategy.

index.html
TallTed:
```suggestion
<b>Process Object</b>. The <code>service</code> query parameter defined in
Section <a href="#did-parameters"></a> identifies a specific `service`
```

index.html
```diff
@@ -2178,19 +2178,133 @@ <h1>Execute Resolution Request</h1>
...
+The object has a <code>type</code> member whose value either:
+											<ul>
+												<li>
+is the string <code>"PathHandlingObject"</code>, or
+												</li>
+												<li>
+is an array that contains the string
+<code>"PathHandlingObject"</code> as one of its elements; and
+												</li>
+											</ul>
```
```suggestion
The object has a <code>type</code> member whose value is one of the following:
											<ul>
												<li>
The string <code>"PathHandlingObject"</code>.
												</li>
												<li>
An array that contains the string <code>"PathHandlingObject"</code> as one
of its elements
												</li>
											</ul>
```

index.html
```diff
@@ -2178,19 +2178,133 @@ <h1>Execute Resolution Request</h1>
...
+										</li>
+										<li>
+The object has a <code>path</code> member.
+										</li>
+									</ul>
+									<p>
+Objects that do not satisfy both conditions MUST be ignored for the purposes of
+path handling strategy determination.
```
```suggestion
Objects that do not satisfy both conditions MUST be ignored for purposes of
determining the path-handling strategy.
```

index.html
```diff
@@ -2178,19 +2178,133 @@ <h1>Execute Resolution Request</h1>
...
+If no Path Handling Objects are found, skip the remaining sub-steps and continue
+the Determine Handling Strategy algorithm.
+								</li>
+								<li>
+Extract the path component from the [=DID URL=] and the value of the
+<code>path</code> member from each collected Path Handling Object.
+								</li>
+								<li>
+For each collected Path Handling Object, compare the [=DID URL=] path against
+the object's path value character by character from the beginning, stopping at
```
```suggestion
the object's path value, character by character, from the beginning, stopping at
```

index.html
```diff
@@ -2178,19 +2178,133 @@ <h1>Execute Resolution Request</h1>
...
+										<li>
+If more than 1 object has the same largest number of matched characters, the
+error `invalidDIDDoc` MUST be returned.
+										</li>
+									</ul>
+								</li>
+								<li>
+Set the handling strategy to <b>Process Object</b>. The strategy MUST use the
+selected Path Handling Object, the [=DID URL=], the [=DID document=], and MAY
+use the [=DID document metadata=].
```
```suggestion
selected Path Handling Object, the [=DID URL=], and the [=DID document=], and
MAY use the [=DID document metadata=].
```

index.html
TallTed: Tweaked @wip-abramson's suggestion. It does seem to be missing something of the original intent. Maybe try using casual language to describe what is desired, and then we can polish that?
```suggestion
<b>DID Object</b>: Collect from the [=DID document=] and [=DID document
metadata=] any objects known to affect dereferencing. If none are found, continue
to the next step. If any are found, use the specifications that define these
objects to determine if any are targeted by the DID URL. If more than one is
targeted, throw an error. The targeted object MUST define a handling strategy.
```

peacekeeper:
I like the general direction of this and how it makes dereferencing more modular and more extensible.

I'm not sure yet if the assumption that we start by determining a single handling strategy really works for all the DID URL dereferencing use cases we have out there. Sometimes, a more complex combination of steps or strategies may be required, not sure, I need to think about this more. E.g. it's not clear to me yet how existing functionalities such as "relativeRef" and "verificationRelationship" fit in.

I don't like the idea of introducing new types or properties, such as a "path" property or a "PathHandlingObject" type. This should be done in extensions. The DID URL Dereferencing algorithm should allow such extensions, but it should not itself define specific types of new objects and how to process them. Are there "path handling objects" other than "services"? And what about extensions that may define ways of dereferencing a path without requiring any "objects"? The way how this is currently written to me feels too narrow and too restrictive.

TallTed (changes requested):
github captured a couple of comments. here they are, a few hours later.

index.html
wip-abramson: I think this is pretty good actually @TallTed.

Basically I was trying to flag that there are extensions that may affect dereferencing, if you encounter any extensions that you understand to affect dereferencing then handle them accordingly. This is the step to do so.

I think the only requirement this spec should place on this is that there should only be one object targetted.

wip-abramson:
I feel like we need a section somewhere in the specification that defines what a "Path Handling Object" is.

@swcurran do you imagine this is going to go in a separate PR?

swcurran:
I'll go through this today for updates and decide on that.  I think it might as well go in this one -- thinking about where it should go.

index.html
swcurran: I put it there because `Identified Object` takes precedence even if a path is present.  If I have a DIDDoc with a  `service` of type `PathService` (or whatever we call it), and in the DID URL I include a path and `?service=<id>` for the object, the algorithm never gets to the Path Handling logic -- the `service` parameter takes priority. But I still need to do the Path Handling implied by the selection of the `service` because of its type.

My previous approach to DID URL handling accounted for the possiblity of the two occurng together via the "with a path", "without a path" handling, with `service` (and the like) dealt with in both.

index.html
swcurran: Change applied

index.html
swcurran: Updated to put in the proper expression of errors -- matching other places in the  spec -- with the caveat that without HTTP binding, its "flexible" what DID URL Dereferencing returns.  That said, even with that "flexibility" -- there needs to be a way to say the type of error, and this is as good as any.

On whether this specific case is an error -- I definitely think it is.  A DIDDoc should not have two Path Handling Objects that result in the same path matching both.  I think we landed on that in previous meetings.

I'll note that a counter that was put forth to that decision was that allowing for multiple matches is a way to allow for the same resource being found in different places.   Although we decided to not do that at the "path matching" layer, I think that can be handled within the selected object processing. For example, in a `service`, the `serviceEndpoint` can be an array. Presumably, if one is selecting that service via a Path, the implication of an array. of endpoints is that the same resource is at each of the path adjusted serviceEndpoints. A specific implementation can decide what to do in that case: get the first, randomly select, get all and compare, or whatever other strategy the developer thinks would be cool.

index.html
swcurran: Applied Ted's changes, but...

I intend that the Handling Strategy "Process Object" (or whatever we call it) means that there is an object from the DIDDoc/Metadata that is to be processed according to its specification and type.  I don't intend that "LinkedResources" and "DID Linked Resources" and "PathService" are each their own Handling Strategy. All of those fall under the category "Process Object", because at the end of this "Determine" sub-process we have identified EXACTLY one object, and the type and specification of the object defines exactly how it will be processed in the next sub-process.  As such,  the last line of Ted's text doesn't really work for me.

But let's talk about...

index.html
swcurran: Much better.

swcurran:
Pushed an update based on the provided feedback.  I've marked the ones that are good cleanups as "resolved", and left unresolved any with open questions, even if I applied the changes.  Hoping we can talk about them when we next cover this PR.

index.html
```diff
@@ -2178,19 +2178,137 @@ <h1>Execute Resolution Request</h1>
...
+and query components of the [=DID URL=], to determine the handling strategy.
+					</p>
+					<p>
+Execute the following steps until a handling strategy is determined. While the
+full [=DID URL=] is available in these steps, query parameters that are not
+applicable to the determination of the handling strategy are ignored.
+					</p>
+					<ol class="algorithm">
+						<li>
+<b>Identified Object</b>: If there are [=DID URL=] query parameters defined in
```
swcurran, line 2198: @msporny raised a concern on the last call about the use of the term "objects".  I've asked Claude about what other terms it would suggest in its place, in hopes that we can have a brief conversation about this either here as replies to this comment, or in an upcoming meeting.

Suggestions:

- Item
- Node
- Entry
- Entity
- Element
- Structure
- Record
- Descriptor

I'm OK with Item, Node (don't know all the JSON-LD implications...), Element.  The others seem pretty bad to me... :-).

Oh and I'm good with "Object".

peacekeeper:
> And what about extensions that may define ways of dereferencing a path without requiring any "objects"? The way how this is currently written to me feels too narrow and too restrictive.

To elaborate on this comment a bit more. There could be extension specifications that define ways of handling specific DID paths, without using any kind of "Path Handling Objects". We even considered at some point adding a section in the [DID Resolution Extensions](https://www.w3.org/TR/did-extensions-resolution/) where such extensions for specific paths could be registered. So for example, one could write a specification for a path `/stats` that would return statistics about the DID. Such an extension specification may define a dereferencing process that wouldn't use any "Path Handling Objects", it would just return some content directly. Correct me if I'm wrong, but as far as I remember this is pretty much how `/whois` was envisioned at some point in the past, i.e. it would be dereferenced consistently across different DID methods using some custom logic.

If I understand correctly, the algorithm proposed in this PR would not support such extensions, since there are only two options for the path: 1. Either you use "Path Handling Objects", or 2. you do something method specific. But it wouldn't support custom extensions for specific paths in the way we have envisioned it before. This is why to me this proposed algorithm feels too narrow and too restrictive. It would be better if DID Resolution didn't restrict innovation on the DID path component at all, and instead leave dereferencing of the DID path completely to DID methods and (registered or unregistered) extensions.

See https://github.com/w3c/did-extensions/issues/615 and the [2025-01-08 Special Topic Call](https://www.w3.org/2025/01/08-did-minutes.html) for more discussion on this.

peacekeeper:
> If I have a DIDDoc with a  `service` of type `PathService` (or whatever we call it), and in the DID URL I include a path and `?service=<id>` for the object, the algorithm never gets to the Path Handling logic -- the `service` parameter takes priority.

This is one of the aspects that have always bothered me about the "Path Handling Object" / "PathService" idea, i.e. the fact that the DID query is processed before the DID path. This is even clearer when we consider `versionId` and `versionTime` (see examples in https://github.com/w3c/did-resolution/issues/332).

E.g. if you have this:

```
did:ex:1234/product/F9f7866e1-4bc5-4495-9f75-308350a37280?versionTime=2020-12-20T19:17:47Z
```

Then assuming the "Path Handling Object" algorithm, the "versionTime" is processed before the DID path. And as Stephen says, if there was a "service" DID parameter, then that would also take precedence over the DID path.

I would argue that this is completely unintuitive for any developer looking at this. In HTTPS URLs, everybody would expect the query to be semantically and/or functionally relative to the path, e.g. in an example like this:

```
https://library.com/book/122?chapter=5
```

The "chapter" parameter would refer to a chapter of a specific book. That parameter would usually be considered semantically and functionally relative to the authority and path of the URI.

I think RFC3986 does NOT actually mandate that the meaning or processing of the query is secondary to the path, instead it says "The path component contains data, usually organized in hierarchical form, that, along with data in the non-hierarchical query component serves to identify a resource". So I guess strictly speaking you could do what you're proposing, i.e. process the query before the path. But it feels very weird to me, and I think it could have all sorts of unexpected side effects.

index.html
msporny: The issue I had was not necessarily with the word "object", it was the capitalization of it and it's use in type names such as `PathHandlingObject`. If you just lowercase "Object" to "object" and rename `PathHandlingObject` to `PathHandler`, that would address my concern.

Many other W3C specs use the INFRA spec to define these words, an "object" is referred to as a "ordered map", or just "map", in that spec: https://infra.spec.whatwg.org/#ordered-map 

It is what we used to define the DID spec: https://www.w3.org/TR/did-1.0/#data-model

We use it in Data Integrity as well to define inputs and outputs to algorithms: https://w3c.github.io/vc-data-integrity/#add-proof

... not that I'm recommending we switch to it.

index.html
```diff
@@ -2178,19 +2178,137 @@ <h1>Execute Resolution Request</h1>
...
+this specification, in the <a>DID method</a> specification, or in a DID
+Extension [[?DID-EXTENSIONS-PROPERTIES]] that identifies a specific object in
+the [=DID document=] or [=DID document metadata=], set the handling strategy to
+<b>Process Object</b>. The <code>service</code> query parameter defined in
```
msporny, line 2202:
```suggestion
the [=DID document=] or [=DID document metadata=], set the handling strategy to
the result of running the algorithm in section [[[#get-strategy]]]</a>. 
The <code>service</code> query parameter defined in
```

msporny: Just as a suggestion for the thing I'm trying to get at... don't take the suggestion literally, just trying to convey how you could change this to remove the way you are referring to other algorithms in the spec. See here for another example:

https://github.com/w3c/vc-data-integrity/blob/main/index.html#L1994

index.html
```diff
@@ -2178,19 +2178,137 @@ <h1>Execute Resolution Request</h1>
...
+										<li>
+The object has a <code>type</code> member whose value is one of the following:
+											<ul>
+												<li>
+The string <code>"PathHandlingObject"</code>, or
+												</li>
+												<li>
+An array that contains the string <code>"PathHandlingObject"</code> as one
+of its elements; and
+												</li>
```
msporny, line 2225:
```suggestion
Collect all path handling objects within the [=DID document=] and [=DID document
metadata=]. A <dfn>path handling object</dfn> is defined as any JSON object that
satisfies both of the following conditions:
									<ul>
										<li>
The object has a <code>type</code> member whose value is one of the following:
											<ul>
												<li>
The string <code>"PathHandler"</code>, or
												</li>
												<li>
An array that contains the string <code>"PathHandler"</code> as one
of its elements; and
												</li>
```

msporny: I literally mean this change. :)

I care about the name of the `type` value... and care much less about what we call it in prose (though, suggest we say `path handling object` or something like that. Capitalizing stuff like "Path Handling Object" adds confusion because it's not clear what you mean by that... for stuff like that, that has a definition, we usually use DFN and then the [= term =] mechanism to refer to the definition, which looks like what you're trying to do here.

I don't really care if the path handling object is DFN'd or not.

index.html
```diff
@@ -2178,19 +2178,137 @@ <h1>Execute Resolution Request</h1>
...
+determining the path-handling strategy.
+									</p>
+								</li>
+								<li>
+If no Path Handling Objects are found, skip the remaining sub-steps and continue
+the Determine Handling Strategy algorithm.
+								</li>
+								<li>
+Extract the path component from the [=DID URL=] and the value of the
+<code>path</code> member from each collected Path Handling Object.
```
msporny, line 2243:
```suggestion
Extract the path component from the [=DID URL=] and the value of the
<code>path</code> member from each collected [=path handling object=].
```

If you do DFN `path handling object`, you can refer to it like this.

swcurran:
From @peacekeeper 's comment [here](https://github.com/w3c/did-resolution/pull/340#issuecomment-4723736862):

> We even considered at some point adding a section in the [DID Resolution Extensions](https://www.w3.org/TR/did-extensions-resolution/) where such extensions for specific paths could be registered. So for example, one could write a specification for a path /stats that would return statistics about the DID. Such an extension specification may define a dereferencing process that wouldn't use any "Path Handling Objects", it would just return some content directly.

I can envision multiple ways to implement this example. A PathHandlingObject could be used for it pointing to a specific path, and the resource statically retrieved or dynamically generated. If the idea was to use some sort of blockchain specific mechanism, the PathHandlingObject would be skipped and land at the DID Method Specific mechanism. I hope there are enough options that a concrete use case could be solved with ingenuity given the guardrails provided. It can handle all of the use cases I've heard of, and that I've seen in the extensions.

From @peacekeeper's comment [here](https://github.com/w3c/did-resolution/pull/340#issuecomment-4723902328)

> This is one of the aspects that have always bothered me about the "Path Handling Object" / "PathService" idea, i.e. the fact that the DID query is processed before the DID path. 

I want to make it clear that while the _determination_ of the handling strategy is ordered, the full DID URL, DID Doc, and DID Doc Metadata is available in executing the strategy. The data has not been thrown away.

index.html
swcurran: Got it.  I'll put this into the next batch of changes.

index.html
jandrieu, line 2202: This excludes other query parameters, making this an exceptional bit of centralization: only those terms in this spec are supported. 

IMO, we should special case Service and PathHandler types, and keep a fallback for the other query parameters.

wip-abramson: > IMO, we should special case Service and PathHandler types, and keep a fallback for the other query parameters.

This confused me on the call a bit. Only `service` a query parameter. I get that you want to special case that, so that if a DID URL contains a `service` query parameter that is used to identify the service object in a DID document to use for path handling.

But I still think clients have to process all other query parameters that they understand to identify identify objects in the DID document before we get to the PathHandler path matching part.

So maybe it is:

1. If `service` present, handle that.
2. Handle any other query parameters you understand to be able identifying path handling objects
3. If no object identified and path present, use path matching algorithm to select path handling object from all objects with PathHandler type in DID Document and DID Document Metadata.
4. Fall back to any other known strategies for path handling
5. Fall back to DID Document strategy.

I am not convinced about special casing `service` but I am fine if we want to do that. Just for me, all the other query parameters cannot guarantee the ordering they are applied in. My preferred approach would be the client applies all query parameters they understand to be relevant and where the DID URL is nonsense resolution fails. We could also say, clients apply query parameters in whatever order they choose.

e.g. `did:example:1234?service=123&linkedResource=456`

For some to be defined query parameter that acts like `service` but for the linkedResource property.

What should happen here? It could be:

- The service 123 is targeted
- An error is raised because both service 123 and linkedResource 456 are targetted
- The client gets to choose which out of service 123 or linkedResource 456 are targetted

wip-abramson:
On today's discussion about **ProcessObject** as a path handling strategy vs the object's type defining the path handling strategy.

I do think the two approaches are closer than they seem, just we are confusing things be using the same language to define different things.

For the record, I do not like ProcessObject as a handling strategy. It is not that, but as an intermediary step I can understand it and also as an output to step 3 in the algorithm can make sense. I don't like the name.

Step 3 could be about selecting the path handling object. 

Step 4 could be about interpreting the type of that object, and if understood, executing its handling stategy.

That is how I think @swcurran has been thinking about it.

Or it step 3 could do the interpretation of the object to identify the handling strategy.

Do we have agreement of the output of step 3 / the input to step 4?

For me it is the DID URL, and options, the path handling object that defines the handling strategy. 

@jandrieu do you see this differently?

peacekeeper:
@swcurran 

> I can envision multiple ways to implement this example. A PathHandlingObject could be used for it pointing to a specific path, and the resource statically retrieved or dynamically generated. If the idea was to use some sort of blockchain specific mechanism, the PathHandlingObject would be skipped and land at the DID Method Specific mechanism. I hope there are enough options that a concrete use case could be solved with ingenuity given the guardrails provided. It can handle all of the use cases I've heard of, and that I've seen in the extensions.

Maybe your PathHandlingObjects could handle many use cases, but still, your approach would introduce limitations (not "guardrails") on DID URL dereferencing that would not support extensions in the way the current algorithm allows them. See the previous discussions the WG had about this topic in https://github.com/w3c/did-extensions/issues/615 and the [2025-01-08 Special Topic Call](https://www.w3.org/2025/01/08-did-minutes.html). Your own `/whois` idea would not be possible, at least not with the [same design](https://identity.foundation/didwebvh/v1.0/#the-whois-use-case) that you originally proposed.

> I want to make it clear that while the _determination_ of the handling strategy is ordered, the full DID URL, DID Doc, and DID Doc Metadata is available in executing the strategy. The data has not been thrown away.

I didn't say the data is thrown away, I'm saying that from a generic URI perspective, it's very unintuitive and problematic to handle the query before the path, and even allow conflicting situations between path and query such as the one you highlighted.

peacekeeper (changes requested):
I'm sorry I have to -1 this too for now. I really like the overall structure and approach involving "handling strategies". But as I said in my comments, I think 1. the "path handling objects" approach is too limiting for potential future extensions, and 2. the fact that the query is handled before the path is really unintuitive and will likely have unexpected side effects.

swcurran:
Per discussion at the DID Working Group Meeting 2026.07.02, closing this PR as its intent is to be included in #344.

Issue #341, opened 2026-06-12
Improve hedgy text around use of DID Parameters


wip-abramson:
We missed one from TAG Review:

> "DID parameters might be used if there is a clear use case where the parameter needs to be part of a URL that references a [resource](https://www.w3.org/TR/did-resolution/#dfn-resources) with more precision than using the [DID](https://www.w3.org/TR/did-resolution/#dfn-decentralized-identifiers) alone." is hedgy ... this should probably be a SHOULD, and I think the note below this is a good description of what should be recommended.

Seems like a straightforward enough change if folks agree to change the Note into normative should text?

wip-abramson:
Text in question here: https://w3c.github.io/did-resolution/#did-parameters:~:text=DID%20parameters%20might%20be

peacekeeper:
I have been wondering about this too. I remember how I originally wrote the text in that green note box, at some point in the past when we were discussing the difference between DID parameters and options. So I ended up with this statement:

> _The important distinction is that DID parameters that are part of the [DID URL](https://www.w3.org/TR/did-core/#dfn-did-urls) should be used to specify what is being identified, whereas options that are not part of the [DID](https://www.w3.org/TR/did-core/#dfn-decentralized-identifiers) or [DID URL](https://www.w3.org/TR/did-core/#dfn-did-urls) should be used to control how resolution and dereferencing are performed._

I have been wondering if people agree with this language, especially in the context of the proposed did->didUrl change, and considering how did:btcr2 and did:cel/scid/keri use DID parameters and options.

ottomorac:
From WG Discussion on 18-Jun: Text to be kept:

[...]

PR #342, opened 2026-06-15
Updates stemming from the DID to DID URL resolution change

swcurran:
A first cut at the changes to the specification stemming from the DID to DID URL change for resolution. This covers text changes -- I didn't look at the pictures.  I used a Claude to generate a list of the places to change, and to work on exact text for some of the updates. I made all the changes to the file.  I will try to do another pass to look for places missed, or would appreciate help from anyone else that identifies additional updates needed.

Some notes:

- Changes of DID --> DID URL and code references from `did` --> `didUrl` where needed. Took care to not do a blanket change, even in the DID resolution algorithm to make sure the instance needed a change.
- General tweaks to wording related to DID to DID URL changes
- Changed the abstract as I was surprised to see it didn't reference DID URL Dereferencing.
- Removed text about "prefering options over parameters"
- Added text about options taking precedence over like named parameters
- Included an example of why you would use both a DID URL and options
- Big Change -- the start of the DID resolution algorithm.  Changed first steps "validate DID" to:
  - Validate DID URL
  - Parse the DID URL to get the "input DID" (eliminates wording changes in later steps) and the parameters as options with the options having precedence.
  - Validate DID
- Clarifcation to use options that are relevant to DID Resolution, and ignore others.
- Did not change the "old" DID URL dereferencing algo, but did update the "new" algorithm
- Added examples of DID URLs passed into DID Resolution.

As noted, help in identifying other places that need to be changed are welcome.

Hopefully this isn't too large a PR -- we'll see :-)

***
<a href="https://pr-preview.s3.amazonaws.com/swcurran/did-resolution/pull/342.html" title="Last updated on Jun 19, 2026, 2:41 PM UTC (1747072)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-resolution/342/9e330e0...swcurran:1747072.html" title="Last updated on Jun 19, 2026, 2:41 PM UTC (1747072)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -161,16 +161,17 @@
   <body data-cite="infra rfc3986 did">
     <section id="abstract">
       <p>
-Decentralized identifier (DID) resolution is the process of obtaining a DID
-document and accompanying metadata for a specific DID. The process takes a DID
-and a set of resolution options as its input and returns a DID document and
-associated metadata about the resolved document and the resolution request. A
-resolved DID document is a set of information which enables cryptographically
-verifiable interactions with the DID subject, including mechanisms such as
-cryptographic public keys. This specification covers the algorithms and
-guidelines to be used for DID resolution and relies on the core DID
-specification, [[[DID]]], which describes the underlying DID architecture in
-full detail.
+Decentralized identifier (<a>DID</a>) resolution is the process of obtaining a
+<a>DID document</a> and accompanying metadata for a given <a>DID URL</a>.
+<a>DID URL dereferencing</a> extends this by using the resolved <a>DID document</a>
+to obtain the specific resource identified by a <a>DID URL</a> — which may be the
+<a>DID document</a> itself, or a referenced external resource. A resolved
+<a>DID document</a> is a set of information that enables cryptographically
+verifiable interactions with the <a>DID subject</a>, through data such as
+cryptographic public keys. This specification covers the algorithms and guidelines
+for <a>DID resolution</a> and <a>DID URL dereferencing</a>, and relies on the core
+DID specification, [[[DID]]], which describes the underlying <a>DID</a> architecture
+in full detail.
       </p>
     </section>
 
@@ -204,7 +205,7 @@ Sporny, Drummond Reed, Joe Andrieu, and Heather Vescent.
 
       <p>
 [=DID resolution=] is the process of obtaining a <a>DID document</a>
-for a given <a>DID</a>. This is one of four required
+for a given <a>DID URL</a>. This is one of four required
 operations that can be performed on any DID ("Read"; the other ones being
 "Create", "Update", and "Deactivate"). The details of these operations differ
 depending on the <a>DID method</a>. Building on top of [=DID
@@ -226,7 +227,7 @@ various security and privacy considerations relevant to implementing a
 
       <p>
 Note that while this specification defines some base-level functionality for DID
-resolution, the actual steps required to communicate with a DID's
+URL resolution, the actual steps required to communicate with a DID's
 <a>verifiable data registry</a> are defined by the applicable
 <a>DID method</a> specification.
       </p>
@@ -234,7 +235,7 @@ resolution, the actual steps required to communicate with a DID's
       <section id="implementer-overview" class="informative">
         <h2>Implementer Overview</h2>
         <p>
-By invoking a <a>DID resolver</a> using the standard `resolve(did,
+By invoking a <a>DID resolver</a> using the standard `resolve(didUrl,
 resolutionOptions)` interface (as defined in the
 <a href="#resolving">DID Resolution section</a>), one can obtain a <a>DID document</a>
 and accompanying metadata (e.g., `contentType`, proof, versioning), which an
@@ -477,16 +478,6 @@ Properties Extensions mechanism [[?DID-EXTENSIONS-PROPERTIES]], to avoid
 collision with other uses of the same DID parameter with different semantics.
       </p>
 
-      <p>
-DID parameters might be used if there is a clear use case where the
-parameter needs to be part of a <a>DID URL</a> that
-references a <a>resource</a> with more precision than
-using the <a>DID</a> alone. It is expected that DID
-parameters are <em>not</em> used if the same functionality can be
-expressed by passing <a href="#did-resolution-options">DID resolution options</a> or
-<a href="#did-url-dereferencing-options">DID URL dereferencing options</a>.
-      </p>
-
       <p class="note" title="DID parameters and DID resolution">
 The <a>DID resolution</a> and the
 <a>DID URL dereferencing</a> functions can be influenced by passing
@@ -494,15 +485,14 @@ The <a>DID resolution</a> and the
 <a href="#did-url-dereferencing-options">DID URL dereferencing options</a> to a
 <a>DID resolver</a> or <a>DID URL dereferencer</a>. Such
 options are not part of the <a>DID</a> or
-<a>DID URL</a>. This is comparable to HTTP, where
-certain parameters could either be included in an HTTP URL, or
-alternatively passed as HTTP headers during the dereferencing process.
-The important distinction is that DID parameters that are part of the
-<a>DID URL</a> should be used to specify
-<em>what is being identified</em>, whereas options that are not part of
-the <a>DID</a> or <a>DID URL</a>
-should be used to control
-<em>how resolution and dereferencing are performed</em>.
+<a>DID URL</a>. For example, a caller might use an unchanged a DID URL used as
+an identifier, and add options to alter the resolution behavior. This is
+comparable to HTTP, where certain parameters could either be included in an HTTP
+URL, or alternatively passed as HTTP headers during the dereferencing process.
+Where same parameter is present in both the
+<a>DID URL</a> and in the <a href="#did-resolution-options">DID resolution options</a>
+or <a href="#did-url-dereferencing-options">DID URL dereferencing options</a>,
+the option value MUST take precedence over the DID URL parameter value. 
       </p>
 
       <section id="datetime">
@@ -664,7 +654,7 @@ ordering with semantic significance.
 
       <p>
 The <a>DID resolution</a> function resolves a
-<a>DID</a> into a <a>DID document</a> by using
+<a>DID URL</a> into a <a>DID document</a> by using
 the "Resolve" operation of the applicable <a>DID method</a> as
 described in <a data-cite="did-core#method-operations">Method Operations</a>.
       </p>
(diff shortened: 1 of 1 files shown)


index.html
swcurran, line 798: I did not make a change here, but I'm wondering if this is the wrong place to process this kind of parameter that alters a DIDDoc after retrieval.  Interested in what others think.

My thinking has been that the DID Resolution step always returns the DIDDoc as it is received from the VDR/DID Method -- _without change_. Parameters such as this that change the DIDDoc are applied at DID URL Dereferencing time, e.g. later in the process. So versionTime/versionId that affect which DIDDoc occur at DID Resolution time, and parameters including `expandRelativeUrls`, `initialState`, `signedIetfJsonPatch` and the one that transforms the keys occur as part of DID URL Dereferencing.

But I can also see the argument that they be applied at DID Resolution time, and the DIDDoc returned for the latter steps in DID URL Dereferencing is the altered DIDDoc.

msporny (approved):

index.html
```diff
@@ -226,15 +227,15 @@ <h1>Introduction</h1>
 
       <p>
 Note that while this specification defines some base-level functionality for DID
-resolution, the actual steps required to communicate with a DID's
+URL resolution, the actual steps required to communicate with a DID's
```
wip-abramson: Do we think it should be DID Resolution still. Or DID URL Resolution?

Noting that here it is DID URL Resolution, but earlier you left is as [=DID Resolution=]

index.html
wip-abramson, line 246: We may want to change this example back to being a DID URL with versionTime. Rather than passing it in options.

index.html
```diff
@@ -477,32 +478,21 @@ <h2>DID Parameters</h2>
...
-should be used to control
-<em>how resolution and dereferencing are performed</em>.
+<a>DID URL</a>. For example, a caller might use an unchanged a DID URL used as
+an identifier, and add options to alter the resolution behavior. This is
+comparable to HTTP, where certain parameters could either be included in an HTTP
+URL, or alternatively passed as HTTP headers during the dereferencing process.
+Where same parameter is present in both the
+<a>DID URL</a> and in the <a href="#did-resolution-options">DID resolution options</a>
+or <a href="#did-url-dereferencing-options">DID URL dereferencing options</a>,
+the option value MUST take precedence over the DID URL parameter value.
```
wip-abramson: Normative MUST should not be in a note. Perhaps just remove the Note part of this and make it a regular paragraph.

Jeffrey suggested this in his TAG review too: https://github.com/w3c/did-resolution/issues/341

index.html
wip-abramson: I think we have an issue tracking this actually - https://github.com/w3c/did-resolution/issues/302

My sense is that resolution should not apply any transformations to the DID document returned by the DID methods algorithm

wip-abramson (approved):
LGTM, only question is about DID Resolution vs DID URL Resolution.

I think DID Resolution is still fine, but either works. We should just pick and be consistent

index.html
swcurran: Thinking more about this, we likely have to take it at least on a case by case basis.  For example, I think `initialState` would have to be done at DID Resolution time, as without it, I think the DIDDoc would be `{}`.

A question is that when the Handling Strategy is not "Return DIDDoc", do the parameters get applied?

wip-abramson: I would say something like `initialState` is consumed by the DID methods resolution algorithm to produce the DID document that is canonical. 
Rather than a transformation that is applied to the canonical DID document.

> A question is that when the Handling Strategy is not "Return DIDDoc", do the parameters get applied?

For me, they would be ignored. The parameters should define the handling strategies they are relevant for.

swcurran: Agreed -- it's a matter of classifying the parameters appropriately as Resolution or Dereferencing DIDDoc. 

In DID Resolution spec parameters:

- Resolution Time: versionTime, versionId, accept
- Dereferencing DIDDoc: None

Extension parameters:

- Resolution Time: initlalState
- Dereferencing DIDDoc Processing: signedIetfJsonPatch, transformKeys

Maybe we should add a note in the section [Query Normailization/Canonicalization by DID Method](https://www.w3.org/TR/did-resolution/#canonicalization-by-did-method) that parameter specifiers are encouraged to identify where in the DID URL Dereferencing algorithm the parameter should be used. A separate PR I think, that might also include DID Extensions Parameters as well as DID Method-specific parameters.

longpd (approved):
I support the PR#342, recognizing that there is more work to be done and additional PRs likely needed. For example, @swcurran makes an interesting case for a PR to address DID Extension Parameters. But overall this is a step forward.

peacekeeper (changes requested):
-1 for the reasons I have articulated, see thread here: https://lists.w3.org/Archives/Public/public-did-wg/2026Jun/0003.html. Several aspects I have brought up have not been sufficiently discussed or were not understood by the group before the vote.

I think this violates the DID WG charter, and would require the PING horizontal review to be re-done.

i will FO to the specification if this goes through.

peacekeeper:
Few additional quick observations, in addition to ones I already mentioned before:
- It makes no sense to first "validate the input DID URL" in step 1, and then "validate the input DID" in step 3. This step 3 will never be triggered.
- This creates confusion about the DID method operations. DID methods define how to create, resolve (previously "read"), update, deactivate DIDs, not DID URLs.
- This PR potentially breaks the HTTPS binding, see my slides for details.

index.html
```diff
@@ -161,16 +161,17 @@
...
-associated metadata about the resolved document and the resolution request. A
-resolved DID document is a set of information which enables cryptographically
-verifiable interactions with the DID subject, including mechanisms such as
-cryptographic public keys. This specification covers the algorithms and
-guidelines to be used for DID resolution and relies on the core DID
-specification, [[[DID]]], which describes the underlying DID architecture in
-full detail.
+Decentralized identifier (<a>DID</a>) resolution is the process of obtaining a
+<a>DID document</a> and accompanying metadata for a given <a>DID URL</a>.
+<a>DID URL dereferencing</a> extends this by using the resolved <a>DID document</a>
```
jandrieu: "extends" feels like the wrong language here.
```suggestion
<a>DID URL dereferencing</a> uses the resolved <a>DID document</a>
```

peacekeeper:
Also, if I read this correctly, this PR would change the algorithms in a way that would pass the fragment to the resolver.

index.html
swcurran: I don't have a strong opinion, but in context I'm leaning to leave it, so have removed the change.

index.html
swcurran: Added an example of using `versionTime` parameter rather than changing it.

index.html
swcurran: Removed the note style from the paragraph. The change had already removed the text before and at the end of the note, as per the discussion at todays DID Working Group call. There is additional text with an example of where an option is added when a DID URL is resolved, and the MUST about options taking precedence over options.

index.html
swcurran: Applied the change.  Thanks.

peacekeeper:
> For example, @swcurran makes an interesting case for a PR to address DID Extension Parameters. But overall this is a step forward.

@longpd Just FYI, extension parameters have always been supported and will continue to be possible in DID URLs, whether this change is approved or not:
https://www.w3.org/TR/did-extensions-resolution/#parameters

Extension parameters are a useful feature, but it is not connected to the change in this PR.

swcurran:
Pushed an update to the PR:

- the changes I commented on above.
- Added text that "a fragment SHOULD NOT be passed in and MUST be ignored".
   - I don't think anyone wants us to error when a fragment is passed in.
   - I did leave it in where I think @peacekeeper noted it, in the parsing step, as if it is passed in, it must be stripped in getting the "base DID".
- removed the "validate DID" step. Totally right, @peacekeeper , it should not be there. I had been wanting to make sure that the "INVALID DID" error had a place. Rather than the unneeded step, I moved it to a possible outcome of DID URL validation.
- Changed the term "input DID" to "base DID" per @jandrieu in DID resolution algorithm section. A separate mini-PR probably needs to be done to both define that term and use it consistently across the spec. For example I did see references to a "DID portion of the DID URL" elsewhere in the spec.

peacekeeper:
> a fragment SHOULD NOT be passed in and MUST be ignored

I think this makes it even worse, as it explicitly acknowledges that this new architecture indeed supports sending fragments to a resolver, which is an idea that (I think) the entire WG has consistently rejected in the last >2 years. Also, it doesn't address the problem of passing the (unused) path to the resolver, which shouldn't happen either.

> didResolutionMetadata: error object with type set to [https://www.w3.org/ns/did#INVALID_DID_URL](https://pr-preview.s3.amazonaws.com/swcurran/did-resolution/pull/342.html#INVALID_DID_URL) or [https://www.w3.org/ns/did#INVALID_DID](https://pr-preview.s3.amazonaws.com/swcurran/did-resolution/pull/342.html#INVALID_DID) **(as appropriate)**

Which one is "appropriate" in what situation, can you give an example? I think you should realize that with this change, there would be no more use for the INVALID_DID error, since there wouldn't be anything anymore in the specs that deal with just DIDs. Everything will always be about full DID URLs everywhere.

> Changed the term "input DID" to "base DID" per @jandrieu in DID resolution algorithm section.

I think the way 4.4 is written now it does the same "comingling" of DID parameters and resolution options that the WG has been criticizing in the last few weeks. If you pass the full DID URL to the resolver, only to then parse it into a "base DID" and options, then you basically have the same design as before the change, except that you have shifted some extra data and processing steps from the client to the resolver.

swcurran:
> I think the way 4.4 is written now it does the same "comingling" of DID parameters and resolution options that the WG has been criticizing in the last few weeks. If you pass the full DID URL to the resolver, only to then parse it into a "base DID" and options, then you basically have the same design as before the change, except that you have shifted some extra data and processing steps from the client to the resolver.

This is why I've always though that the whole DID/DID URL argument is of little consequence. The processing of the "string" has to occur and it is really a question of where. It has taken up a lot of time in the Working Group.

swcurran:
> I think this makes it even worse, as it explicitly acknowledges that this new architecture indeed supports sending fragments to a resolver, which is an idea that (I think) the entire WG has consistently rejected in the last >2 years. Also, it doesn't address the problem of passing the (unused) path to the resolver, which shouldn't happen either.

@peacekeeper -- I would appreciate it if you could lower the rhetoric in your comments and provide concrete suggestions.

I think you are meaning that you disgree that a fragment MUST be ignored, but instead that it should trigger an error, and you think it would be good to add path to the SHOULD NOT be passed and trigger an error.  Is that corrrect?

Feedback that way makes it easier for the the Working Group to better understand what you mean and how to discuss it in a concise way. Even better would be to put that feedback in for form a change so that it could be discussed in GitHub in context.

peacekeeper:
> @peacekeeper -- I would appreciate it if you could lower the rhetoric in your comments and provide concrete suggestions.

Sorry if my language was inappropriate, was it really? I just think this PR in its current form combines the worst aspects of both approaches, since it increases tracking risks by sending unnecessary information to the resolver (path, fragment), and at the same time it still performs the "comingling" of DID parameters and options which (I think) motivated the change of inputs in the first place.

I don't know how to say this in a nicer way. And I think I have very clearly articulated my suggestion, which is to not send full DID URLs to resolvers.

wip-abramson:
@peacekeeper I do not think your language was inappropriate for what its worth. 

However, I can definitely understand @swcurran's frustration. The DID WG finds itself in quite tangled up with respects to the DID Resolution spec at the moment. I appreciate all the time and effort that you Stephen have put in to try and get us unstuck with the many PRs attempts and iterations we have gone through. Only to have them be repeatedly hit roadblocks. 

On Thursday's call I will be presenting an overview of our remaining timeline left under this charter and some realistic options for us to move forwards. 

The worst thing that we can do as a group is not get into CR before the charter runs out. That would be a fail and a huge detriment to the work and the community of implementers that depend on it.

I know we all want what is best for the decentralized identity community and I thank you for continuing to engage in good faith through all the frustrations.

swcurran:
Closing per proposal in DID Working Group Meeting 2026.07.02.

PR #348, opened 2026-07-12
Transition to cr

jandrieu:
This adds a snapshot at /transitions/2026/CR

I set this to v1 in the body of the document.

You may preview at https://legreq.github.io/did-resolution/transitions/2026/CR/index.html

We may want to add a "retiredDate" for Dmitri. FWIW, I did move Markus and Dmitri to "formerEditors" which is a formal respect config value, but it didn't seem to change anything.

Otherwise, this passes all of the respec checks. 

Also, the publication date is tomorrow. Not sure what that *SHOULD* be, but we can change that in the config.

My only concern is that the echidna set up may not be working quite as expected. At least not as I expected.

I had thought there was a data in the TR URL for "current draft", but maybe I'm misunderstanding how that works. Otherwise, we might need to update the config to support that.

@pchampin Can you check that the respec config is correct?

***
<a href="https://pr-preview.s3.amazonaws.com/LegReq/did-resolution/pull/348.html" title="Last updated on Jul 16, 2026, 3:51 PM UTC (613536f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-resolution/348/bed3589...LegReq:613536f.html" title="Last updated on Jul 16, 2026, 3:51 PM UTC (613536f)">Diff</a>


swcurran:
I notice that the index.html file in the transitions/2026/CR folder is about twice the size of the one in the root (337k vs. 153k).  Is that relevant?  I don't know the process for creating the transitions document, but thought it a bit odd. Is it because it is statically generated vs. dynamically for what is in the root?

peacekeeper:
I don't really understand why the entire section "8. DID Resolution Architectures" is marked at-risk? Why not just mark the sub-sub section about "Proxied Resolution", if that's what the WG considers problematic? I also disagree that it "is likely to be replaced with content from the the DID Resolution Threat Model."

I don't understand either why section "10. DID URL Dereferencing Result" is also marked at-risk. This is just a serialization of a DID URL dereferencing result, independent of how the algorithm works?

I admit I don't fully understand the W3C process, but I'm worried that this PR is basically a step towards removing signficant portions of the spec that are perfectly fine.

wip-abramson (approved):
LGTM

peacekeeper (approved):
I still don't really understand why the entire section "8. DID Resolution Architectures" is marked at-risk (discussion seems to be mostly about the proxied resolution sub-sub-section). But I'm still fine with merging.

ottomorac (approved):
Approved during DID WG meeting on 16-July.

mccown (approved):
These changes represent the discussion held in the WG call today (16 Jul 2026) and I concur with these changes.

PR #344, opened 2026-06-30
full pass at DID URL dereferencing algorithm (isolated)

jandrieu:
Per our last working group call (that I was on). Here's a proposed top-level dereferencing algorithm with flow chart illustration.

***
<a href="https://pr-preview.s3.amazonaws.com/LegReq/did-resolution/pull/344.html" title="Last updated on Aug 13, 2026, 3:01 PM UTC (e623427)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-resolution/344/ab715a7...LegReq:e623427.html" title="Last updated on Aug 13, 2026, 3:01 PM UTC (e623427)">Diff</a>

PR #243, opened 2025-10-30
Define network based conformance classes for resolver and dereferencer

wip-abramson:
This attempts to address #213 

I have addressed this issue a little differently than discussed. Specifically around naming of the conformance class.

I have a conforming resolver (the software bit) then a conforming network-based resolver is a conforming resolver that implements HTTP Binding.

I also added a similar class for DID URL dereferencers.

Then I attempted to clarify the requirement for JSON serialization of inputs and outputs for both resolving and dereferencing. 

Reviews appreciated.

***
<a href="https://pr-preview.s3.amazonaws.com/wip-abramson/did-resolution/pull/243.html" title="Last updated on Jan 11, 2026, 8:21 PM UTC (4d40db3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-resolution/243/a0da1e3...wip-abramson:4d40db3.html" title="Last updated on Jan 11, 2026, 8:21 PM UTC (4d40db3)">Diff</a>

PR #356, opened 2026-08-19
Explain better that function signatures are abstract

peacekeeper:
In discussions about the DID URL dereferencing algorithm, I keep hearing statements such as:

- _"a browser should not need to expose an endpoint"_
- _"when I dereference a DID URL, I don't need to implement the exact function in the spec, I just implement an algorithm"_
- _"dereferencing should not require a separate library or SDK"_

I fully agree with all these statements, and I think so does everyone else in the WG. Unfortunately, it seems the current DID URL dereferencing algorithm is sometimes misunderstood as requiring any of the above things, which has never been the intention.

I think the source of some of this misunderstanding may be language which says "implementations do not alter the signature of these functions", which currently exists in both the DID Resolution and DID URL Dereferencing sections.

I traced this down to the following commits some years ago:
- https://github.com/w3c/did/commit/78ffa498ffd39aa93da6b0b45c3f2a33ca10528e
- https://github.com/w3c/did/commit/8bfe90e72893f0ab0e8e1bce0dc5ad6b896aece4
- https://github.com/w3c/did/commit/2dc94f3984ef5b2c44dc80173abe06422adace79

It seems this language was a mistake, so here's a PR which emphasizes that `resolve()` and `dereference()` are abstract interfaces for the specified algorithms with logical inputs and outputs, but there is no requirement to implement them in any specific form.

So for example, when you want to verify a proof on a Verifiable Credential and you dereference a DID URL to a verification method, your implementation of the DID Resolution and DID URL Dereferencing algorithms can take any form you want, they don't have to be implemented as a function call, or SDK, or remote API. To me, this has always been clear anyway, but I can see how the current spec language didn't emphasize this sufficiently.

^ @jandrieu

***
<a href="https://pr-preview.s3.amazonaws.com/peacekeeper/did-resolution/pull/356.html" title="Last updated on Aug 19, 2026, 4:38 PM UTC (371aa9f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-resolution/356/dfb2b95...peacekeeper:371aa9f.html" title="Last updated on Aug 19, 2026, 4:38 PM UTC (371aa9f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -753,13 +753,24 @@ All [=conforming DID resolvers=] MUST implement the
       </p>
 
       <p>
-Conforming <a>DID resolver</a> implementations do not alter
-the signature of this function in any way.
+The signature of this function is to be understood as an abstract interface
+for the specified algorithm, with logical named inputs and outputs. Conforming
+<a>DID resolution</a> implementations are not required to
+implement this function using the same names of the inputs, outputs,
+or of the function itself, nor do they need to implement the algorithm in
+any specific form, such as an explicit function call in a programming language,
+a software library, or a protocol endpoint.
+      </p>
+      <p>
+<a>Bindings</a> such as the <a href="#bindings-https">HTTP(S) binding</a>
+can be used to define specific ways of implementing and exposing this function.
+      </p>
+      <p>
 <a>DID resolver</a> implementations might map the
 <code>resolve</code> function to a method-specific internal function to
 perform the actual <a>DID resolution</a> process.
 <a>DID resolver</a> implementations might implement and
-expose additional functions with different signatures in addition to the
+expose additional functions with different signatures than the
 <code>resolve</code> function specified here.
       </p>
 
@@ -1464,9 +1475,7 @@ dereference(didUrl, dereferenceOptions) →
 
       <p>
 All [=conforming DID URL dereferencers=] MUST implement the [=DID URL
-dereferencing=] function for at least one [=DID method=]. All inputs and
-outputs of the <code>dereference</code>
-function apart from [=contentStream=] MUST be serializable to JSON.
+dereferencing=] function for at least one [=DID method=].
       </p>
 
       <p>
@@ -1556,13 +1565,24 @@ structure is further defined in <a href="#did-url-content-metadata"></a>.
       </dl>
 
       <p>
-Conforming <a>DID URL dereferencing</a> implementations do not alter
-the signature of these functions in any way.
+The signature of this function is to be understood as an abstract interface
+for the specified algorithm, with logical named inputs and outputs. Conforming
+<a>DID URL dereferencing</a> implementations are not required to
+implement this function using the same names of the inputs, outputs,
+or of the function itself, nor do they need to implement the algorithm in
+any specific form, such as an explicit function call in a programming language,
+a software library, or a protocol endpoint.
+      </p>
+      <p>
+<a>Bindings</a> such as the <a href="#bindings-https">HTTP(S) binding</a>
+can be used to define specific ways of implementing and exposing this function.
+      </p>
+      <p>
 <a>DID URL dereferencing</a> implementations might map the
 <code>dereference</code> function to a method-specific internal function
 to perform the actual <a>DID URL dereferencing</a> process.
 <a>DID URL dereferencing</a> implementations might implement and
-expose additional functions with different signatures in addition to the
+expose additional functions with different signatures than the
 <code>dereference</code> function specified here.
       </p>
 


peacekeeper:
In addition, this PR also removes a statement which says:

> All inputs and outputs of the <code>dereference</code> function apart from [=contentStream=] MUST be serializable to JSON.

This sentence was originally added for both DID resolution and DID URL dereferencing, and then in the same PR it was removed only from DID resolution. I think it was simply an oversight that it wasn't removed from DID URL dereferencing as well.

See:
- https://github.com/w3c/did-resolution/pull/243
- Commits https://github.com/w3c/did-resolution/commit/fbde41e68969cfb26b316a90ad6a23edfb639073 and https://github.com/w3c/did-resolution/pull/243/changes/4d40db3b3d84ea9b723dd7a916b7a150f5b6e05c

^ @wip-abramson

ottomorac:
Thanks for the PR Markus. We will hold for now until the discussion about #344 is finished, before we return to discuss this one.

peacekeeper:
@ottomorac I was hoping that small PRs like this (about 40 changed lines, vs. about 600 changed lines in #344) would be easy to review and agree to. I also think this here is just a small clarification that doesn't conflict with anything else that's going on.

index.html
```diff
@@ -753,13 +753,24 @@ <h1>DID Resolution</h1>
...
       <p>
-Conforming <a>DID resolver</a> implementations do not alter
-the signature of this function in any way.
+The signature of this function is to be understood as an abstract interface
+for the specified algorithm, with logical named inputs and outputs. Conforming
+<a>DID resolution</a> implementations are not required to
+implement this function using the same names of the inputs, outputs,
+or of the function itself, nor do they need to implement the algorithm in
+any specific form, such as an explicit function call in a programming language,
+a software library, or a protocol endpoint.
```
wip-abramson, line 762: I actually think this is not correct.

The spec IS trying to define an explicit function signature for DID Resolution - `resolve(did, resolutionOptions)`. This function signature MUST be implemented by conformant DID Resolvers. All inputs and outputs of the function signature MUST be serializable to JSON.

This is the concrete, standard interface that supports interoperability amongst DID methods.

I believe @msporny suggested adding the MUST be serializable to JSON bit, because that is how we test it.

The point of contention in the group is whether this approach applies to DID URL dereferencing. Is dereferencing intentended to be a point of interoperability with standard inputs and outputs, or an algorithm executed by after receiving a standardized DID resolution result.

peacekeeper: @wip-abramson You removed the "must be serializable to JSON" language for DID resolution in https://github.com/w3c/did-resolution/commit/4d40db3b3d84ea9b723dd7a916b7a150f5b6e05c, but you didn't remove it for DID URL dereferencing. Any reason why it shouldn't be removed there as well?

peacekeeper:
@wip-abramson Regarding

> This function signature MUST be implemented by conformant DID Resolvers.

This isn't really how I have been thinking about it. The same point which @jandrieu has been making about dereferencing that I agree with ("you don't have to implement the exact function signature") applies equally to resolution as well in my mind. If you want to resolve a `did:key` using an entirely local process, I don't think you have to implement the exact `resolve(did, resolutionOptions)` function signature in order to be conformant.

The function signatures for both resolve and dereference have always been and should remain abstract, with logical inputs and outputs, that can take different forms when implemented.

> This is the concrete, standard interface that supports interoperability amongst DID methods.

I think the concrete interoperability comes from the HTTP(S) binding, not from forcing every implementation of a resolver to implement an specific function signature with exact names of the inputs, outputs, and the function itself.

TallTed:
> "you don't have to implement the exact function signature"

That's an error, somewhere.

You *do* have to implement the exact function signature, because, as [noted earlier](https://github.com/w3c/did-resolution/pull/356#discussion_r3854479116), that is the facet of interoperability.

You *don't* have to implement the exact algorithm(s) we document, because these are black boxes. If your algorithm takes a given input conforming to the function signature we specify, and you produce the exact output we specify, then your algorithm/implementation is conformant to our specification. This allows for improvement and innovation by implementers without repeating the W3C REC cycle.

index.html
wip-abramson: Sorry @peacekeeper I missed this. The reason I removed this statement was because it was a duplicate, this was identified as part of developing the test suite for DID Resolution. 

I likely did not remove the statement for DID URL Dereferencing because either it was not a duplicate, or I did not include statements from DID URL Dereferencing in my analysis because I was not developing a test suite for them at that time. I cannot remember which.

peacekeeper: Thanks @wip-abramson . I imagine that it was an oversight, and if that statement is removed for DID Resolution, then probably it should also be removed for DID URL Dereferencing.

peacekeeper: Or alternatively, maybe the "MUST be serializable to JSON" part should be added back to DID Resolution? Not sure.

peacekeeper:
> The spec IS trying to define an explicit function signature for DID Resolution - resolve(did, resolutionOptions). This function signature MUST be implemented by conformant DID Resolvers.

> You do have to implement the exact function signature, because, as https://github.com/w3c/did-resolution/pull/356#discussion_r3854479116, that is the facet of interoperability.

I didn't expect this feedback, I thought it was clear that resolvers don't have to implement the exact `resolve()` function signature, and that (local and other) bindings and implementations can have whatever concrete function signatures they want, as long as the logical inputs and results match the abstract `resolve()` function.

Created https://github.com/w3c/did-resolution/issues/363.

Issue #355, opened 2026-08-13
Registered and method-defined error types have no HTTP status mapping


danpape:
did:btcr2 defines its own error types within our spec: INVALID_DID_UPDATE, LATE_PUBLISHING and MISSING_UPDATE_DATA (https://dcdpr.github.io/did-btcr2/errors.html).

Properly defining our own error types looks settled: mint a URL of our own for the type values and register the errors in the error registry in DID Resolution Extensions.

What doesn't seem settled is what an HTTP(S) binding should do with our new error URLs. The error status code table in the HTTP(S) Binding section is keyed on the nine `https://www.w3.org/ns/did#` URIs, so anything else falls to `(any other error URI)` -> 500.

This doesn't seem to be only a DID-method concern. Two classes of error the spec already knows about would be lumped into "(any other error URI)":

- The seven DIF-defined codes registered in DID Resolution Extensions (`notAllowedVerificationMethodType`, `notAllowedKeyType`, and the rest) have no row in the status code table.
- The "Errors" section of the spec states it "uses some errors defined in Section 3.5 Processing Errors of the [CID] specification"; those have no row either.

One specific example of mine: btcr2's MISSING_UPDATE_DATA means the DID is valid and a document existed, but the resolver cannot obtain the data needed to produce its current state. Our method might return, for example, an error with type "https://btcr2.dev/ns/error#MISSING_UPDATE_DATA". That could be considered a NOT_FOUND error, but a conforming HTTP(S) binding must answer 500, which seems incorrect: the client did nothing wrong, and nothing on the server broke.

I realize that catch-all row isn't an oversight. It was proposed in #72 as a default. I'm asking whether it should also be the answer for error types that are registered.

If the WG agrees, I could create a PR something like: 

In the Errors section:

> A specification that defines an error type not defined in this section
> SHOULD declare which of the error types defined in this section it
> refines, and SHOULD register the error type in
> [DID-EXTENSIONS-RESOLUTION].

and in the HTTP(S) Binding section, after the table:

> If an error `type` URI does not appear in the table above and the
> specification defining it declares that it refines an error type that
> does, the HTTP status code MUST be that of the refined error type.
> Otherwise, the HTTP status code MUST be `500`.

I know v1 is at CR now, so a new requirement in the binding may be more than the WG wants to take on right now. I thought I'd still bring it up now because we're working toward being one of the independent implementations the CR calls for, which is how we came to be reading the binding this closely.

peacekeeper:
[...]

Issue #72, opened 2022-08-03
methodNotSupported has no associated HTTP status code


aljones15:
The spec states to throw the error and add it to the `didResolutionMetadata`, but unlike other errors we don't know the status code for it.

peacekeeper:
I agree, this is missing. Do you have a suggestion? Maybe 400?

We should probably also define a "default" HTTP status code (400? 500?)  if the didResolutionMetadata contains any other error that is not explicitly mapped to an HTTP status code.

aljones15:
Those are both issues I've been dealing with this week. I think a default of 500 sounds about right.

One possibility is this for `methodNotSupported`:

[406 Not Acceptable](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406)
This response is sent when the web server, after performing [server-driven content negotiation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation#server-driven_negotiation), doesn't find any content that conforms to the criteria given by the user agent.

As we were unable to find a driver that could resolve the did method given.

peacekeeper:
I agree with 500 as default.

Regarding "methodNotSupported", I don't think 406 is right, since that's for content negotiation, and we're already mapping "representationNotSupported" to 406.

I think "methodNotSupported" should map to either 404 (since the DID is essentially "not found" by that resolver; kind of like asking for a path on a web server that doesn't exist), or 400 (meaning the client sent a bad request).

aljones15:
`404` sounds better to me, but I'm not exactly an expect in this field.

TallTed:
4xx don't feel right, as the user/client did nothing wrong in their request.  Rather, it's the server that doesn't know how to handle the client's properly-formed request -- and thus I think [`501`](https://www.rfc-editor.org/rfc/rfc9110.html#name-501-not-implemented) is the most correct.

> The `501 (Not Implemented)` status code indicates that _the server does not support the functionality required to fulfill the request_. 

Emphasis added in the quotation above.

aljones15:
@peacekeeper `501` is not used currently and seems pertinent.

peacekeeper:
I agree with `501`. Could you maybe review https://github.com/w3c-ccg/did-resolution/pull/73 and https://github.com/w3c-ccg/did-resolution/pull/74 ?

aljones15:
closing as methodNotSupported is now associated with 501.

PR #358, opened 2026-08-26
Map error types not defined in this specification to HTTP status codes

danpape:
Addresses #355.

Per @peacekeeper's suggestion there, a custom error code could either declare which existing error type it refines and inherit that type's status, or declare its HTTP status code directly. Both routes are here.

I made two edits: In the Errors section, a specification that defines an error type not defined there MAY declare which of the defined types it refines. In the HTTP(S) Binding section, a rule after the status code table: a status declared by the defining specification wins, otherwise the status of the type it declares it refines, otherwise `500`.

These fix a gap described in #355. The Errors section says it "uses some errors defined in Section 3.5 Processing Errors of the [CID] specification", but none of those have a row in the status code table. The error registry in DID Resolution Extensions carries seven DIF-defined codes (`notAllowedVerificationMethodType` and the rest) and none of those have a row either. All of them would end up as `(any other error URI)` -> `500` today.

One question: I annotated the catch-all row as `(any other error URI; see below)` rather than deleting it. Step 3 in the text I added kind of makes that row redundant, but the `500` default was proposed deliberately in #72 and I would rather not remove it uninvited.

Two things I can't fix here. The error registry doesn't have a place to put such a declaration as an entry records only where the error is normatively defined, and its example is `{"error": "invalidDid"}`, a bare string rather than an RFC 9457 object with a URL `type`. I've raised that as #359, which is really the broader question of how a registered error type gets a URL at all. And the test suite asserts the exact standard type URI rather than just the status, so a conformant resolver raising a correctly-mapped method error would fail it. I brought that up informally in w3c/did-resolution-test-suite#9.

***
<a href="https://pr-preview.s3.amazonaws.com/danpape/did-resolution/pull/358.html" title="Last updated on Aug 28, 2026, 11:17 PM UTC (1270d0e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/did-resolution/358/d7fc8d2...danpape:1270d0e.html" title="Last updated on Aug 28, 2026, 11:17 PM UTC (1270d0e)">Diff</a>


wip-abramson (approved):
Thanks Dan, I will aim to get this on the agenda next week to get some more attention on it. But it looks good to me.

danpape:
Thanks!

No preference came back on the wording (discussed in the original issue), so I'll leave "refines".

peacekeeper (approved):
Looks good, thanks!

index.html
```diff
@@ -3138,6 +3138,13 @@ <h1>Errors</h1>
         </li>
       </ul>
 
+      <p>
+A specification that defines an error type not defined in this section MAY
+declare which of the error types defined in this section it refines. A binding
+defined in Section <a href="#bindings"></a> can use such a declaration to
+determine how to convey that error type.
```
```suggestion
declare that it refines one of the error types defined in this section.
A binding defined in Section <a href="#bindings"></a> can use such a
declaration to determine how to convey that error type.
```

